PR #2, opened 2020-01-12
Fix patch.js hack script using the wrong arg index

hsjoberg:
Hi!
Because of a breaking change in Deno, `args[0]` no longer contains the script filename, thus the first option is in index 0 instead of 1.

https://github.com/denoland/deno/pull/3628


dyedgreen:
LGTM

PR #16, opened 2020-01-27
Implement Rows#columns

### LINKED ISSUE #15: Support for getting column names from the result of a select query
Hi, thanks for the module at first!

I need a way to get column names from the result of a select query.

## Why?

I'd like to convert result rows into an array of object as follows:

```typescript
  const db = new DB();
  db.query("CREATE TABLE test (id INTEGER, name TEXT)");
  db.query("INSERT INTO test (id, name) VALUES (?, ?)", [1, "name"]);
  const rows = db.query("SELECT id as test_id, name as test_name from test");
  const columns = rows.columns(); // This is provisional code.
  assertEquals(columns, ["test_id", "test_name"]);

  const results = [];
  for (const row of rows) {
    const result = {};
    for (let i = 0; i < row.length; i++) {
      result[columns[i]] = row[i];
    }
    results.push(result);
  }
```

Thanks.

uki00a:
This PR closes #15.

My C skill is poor and I'm not sure that API is correct or not, so I'm very open to comments/suggestions.

Thanks.

## Note

I changed the `build/sqlite.js` as follows:

```
$ cd ./build
$ make release
$ base64 sqlite.wasm --wrap=0 > base64.txt
$ vim sqlite.js
# Edit `wasmBase64` variable
```


build/src/wrapper.c
```diff
@@ -311,3 +311,12 @@ int KEEPALIVE column_bytes(int entry_id, int stmt_id, int col) {
   }
   return sqlite3_column_bytes(stmt, col);
 }
+
+const char* KEEPALIVE column_name(int entry_id, int stmt_id, int col) {
```
dyedgreen: This is very good! Could you add two more wrappers for `sqlite3_column_origin_name` and `sqlite3_column_table_name` in the same manner as this one?

(And make sure to re-base on master before you compile, as there is now support for automatically updating the `sqlite.js` file as part of the `make` invocation 😉)

src/rows.js
```diff
@@ -68,6 +68,24 @@ export class Rows {
...
   }
 
+  /**
+   * Rows.columns
+   *
+   * Call this if you need column names from the result of a select query.
+   */
+  columns() {
+    if (this._done) {
+      return [];
```
dyedgreen: I feel it might be more prudent to raise an error in this case, as we don't truly know what the columns are.
```suggestion
      throw new SqliteError("Unable to retrieve column names as transaction is finalised.");
```

(be sure to import `SqliteError`)

src/rows.js
```diff
@@ -68,6 +68,24 @@ export class Rows {
     return { value: row, done: false };
   }
 
+  /**
+   * Rows.columns
+   *
+   * Call this if you need column names from the result of a select query.
```
dyedgreen: Add a comment about the returned objects shape (i.e. that it exposes name, originName, and tableName) and maybe mention which sqlite functions they correspond to (e.g. in a table).

src/rows.js
```diff
@@ -68,6 +68,24 @@ export class Rows {
...
+   */
+  columns() {
+    if (this._done) {
+      return [];
+    }
+
+    const columnCount = this._db._wasm.column_count(this._db._id, this._id);
+    const columns = [];
+    for (let i = 0; i < columnCount; i++) {
+      columns.push(getStr(this._db._wasm, this._db._wasm.column_name(this._db._id, this._id, i)));
```
dyedgreen: Please create `{name, originName, tableName}` objects as-per our API discussion earlier 😄

test.ts
```diff
@@ -447,6 +447,18 @@ test(function invalidBindDoesNotLeakStatements() {
   db.close();
 });
 
+test(function getColumnsFromRows() {
+  const db = new DB();
+
+  db.query("CREATE TABLE test (id INTEGER, name TEXT)");
+  db.query("INSERT INTO test (id, name) VALUES (?, ?)", [1, "name"]);
```
dyedgreen: Nit: If you make id a `INTEGER PRIMARY KEY AUTOINCREMENT`, you won't have to set them here. But this is fine too (I think it might actually be shorter, as you only add one key 😅)

uki00a:
@dyedgreen 
Thanks for comments!
I have fixed the issues you pointed out.
Could you review it again, please?

build/Makefile
```diff
@@ -24,7 +24,7 @@ SQLFLG = -DSQLITE_DQS=0 -DSQLITE_THREADSAFE=0 -DSQLITE_LIKE_DOESNT_MATCH_BLOBS\
          -DSQLITE_DEFAULT_FOREIGN_KEYS=1 -DSQLITE_TEMP_STORE=3 -DSQLITE_DISABLE_LFS\
          -DSQLITE_OMIT_DEPRECATED -DSQLITE_OMIT_UTF16 -DSQLITE_OMIT_SHARED_CACHE\
          -DSQLITE_OMIT_LOAD_EXTENSION -DSQLITE_OMIT_PROGRESS_CALLBACK -DSQLITE_OMIT_TRACE\
-         -DSQLITE_OS_OTHER=1 -DSQLITE_OMIT_COMPLETE -DSQLITE_OMIT_WAL
+         -DSQLITE_OS_OTHER=1 -DSQLITE_OMIT_COMPLETE -DSQLITE_OMIT_WAL -DSQLITE_ENABLE_COLUMN_METADATA
```
dyedgreen: Can you add a line in the `Rational' comment explaining why this flag is added?

(also, I'm curios if it effects benchmarks negatively, I'll have a look at that...)

src/rows.js
```diff
@@ -68,6 +69,35 @@ export class Rows {
     return { value: row, done: false };
   }
 
+  /**
+   * Rows.columns
+   *
+   * Call this if you need column names from the result of a select query.
+   *
+   * This method returns an array of objects, where each object has the following properties.
```
dyedgreen: nit: punctuation:
```suggestion
   * This method returns an array of objects, where each object has the following properties:
```

src/rows.js
```diff
@@ -68,6 +69,35 @@ export class Rows {
...
+   *
+   * | Property     | Value                                      |
+   * |--------------|--------------------------------------------|
+   * | `name`       | the result of `sqlite3_column_name`        |
+   * | `originName` | the result of `sqlite3_column_origin_name` |
+   * | `tableName`  | the result of `sqlite3_column_table_name`  |
+   */
+  columns() {
+    if (this._done) {
+      throw new SqliteError("Unable to retrieve column names as transaction is finalised.")
```
dyedgreen: Nit: can we spell this `finalized` to reflect the SQLite3 spelling in `sqlite3_finalize`? (s is not technically wrong, but I'd like to be consistent with the SQLite3 docs)

test.ts
```diff
@@ -447,6 +447,28 @@ test(function invalidBindDoesNotLeakStatements() {
   db.close();
 });
 
+test(function getColumnsFromRows() {
```
dyedgreen: Can you add a test for the case where the name is not specified (using an `column AS name`), so that we notice if SQLite changes it's behaviour (or at least have a chance to notice 😅)

dyedgreen:
### re benchmarks:

I'll check if there is a significant performance impact when enabling the column-names vs when they are disabled once I have time today or over the weekend. If they prove very expensive, we might consider not including the origin information (although I would quite like to provide these in the API, so we'll see how it goes.)

Apologies for not bringing this up earlier; I only read the SQLite3 docs hastily and didn't see the bit about the extra flag 😅

dyedgreen:
re benchmarks:

Python does not seem to expose these fields actually. See: https://docs.python.org/3/library/sqlite3.html#accessing-columns-by-name-instead-of-by-index

build/Makefile
uki00a: https://www.sqlite.org/c3ref/column_database_name.html

> These APIs are only available if the library was compiled with the SQLITE_ENABLE_COLUMN_METADATA C-preprocessor symbol.

dyedgreen:
### re benchmarks:
Just checked on my machine, and there does not seem to be a significant difference (your CI run seems to have had some bad luck; sorry for stirring up all this drama 😅)

docs/api.md
```diff
@@ -214,6 +214,31 @@ Always use `.done()` instead of `break`.
...
+
+This method returns an array of objects, where each object has the following properties.
+
+| Property     | Value                                      |
+|--------------|--------------------------------------------|
+| `name`       | the result of `sqlite3_column_name`        |
+| `originName` | the result of `sqlite3_column_origin_name` |
+| `tableName`  | the result of `sqlite3_column_table_name`  |
+
+```javascript
```
dyedgreen: Two things here:
1. This file is auto-generated and your changes here will be discarded with re-generation (which will happen automatically on push to master)
2. I don't think we need the code example here, the documentation you provide in the documenting comment is completely sufficient, explaining what the returned objects will look like 😃

dyedgreen: (to fix this, simply run `deno --allow-read --allow-write docs/generate.js` and follow the instructions given)

src/rows.js
```diff
@@ -68,6 +69,35 @@ export class Rows {
...
+   *
+   * | Property     | Value                                      |
+   * |--------------|--------------------------------------------|
+   * | `name`       | the result of `sqlite3_column_name`        |
+   * | `originName` | the result of `sqlite3_column_origin_name` |
+   * | `tableName`  | the result of `sqlite3_column_table_name`  |
+   */
+  columns() {
+    if (this._done) {
+      throw new SqliteError("Unable to retrieve column names as transaction is finalized.")
```
dyedgreen: nit: missing semicolon on end of line

uki00a:
@dyedgreen @hsjoberg 

Thanks!

PR #19, opened 2020-02-01
Build sqlite library from src

### LINKED ISSUE #18: Add `SQLITE_OMIT_ATTACH ` to build and add documentation
I think omitting the `ATTACH` keyword would be a wise move (since the VFS does not really support different file-names). This means that people could make the following mistake:
```SQL
ATTACH DATABASE db1 as db1;
ATTACH DATABASE db2 as db2;
```
Now `db1` and `db2` point to the same database as the VFS only inspects the first letter of the name. Further, the files would point to the journal file, as journal files are recognised using the length of the file name.

dyedgreen:
Build sqlite3 from canonical sources to ensure `SQLITE3_OMIT_*` work correctly:

> **Important Note:** The SQLITE_OMIT_* options may not work with the amalgamation. SQLITE_OMIT_* compile-time options usually work correctly only when SQLite is built from canonical source files.

Closes #18


dyedgreen:
Considering to not include SQLite source, but rather download it if needed... It does add quite a bit of files to the repo

PR #30, opened 2020-03-07
Add CurrentTime to VFS

### LINKED ISSUE #29: Need for SQLite date and time functions
I'm making good progress on a deno/sqlite project, but I need the sqlite date and time functions for me to proceed. For example the following query

    select datetime('now', 'localtime') as "start_time"

should return the current timestamp, but it returns a value of null. See [this page](https://sqlite.org/lang_datefunc.html) for details.

I need that before I can continue with my project, but I also found the date and time constants for defaults don't work, eg

    create table foo (as_of text default current_date)

or

    insert into foo (as_of) values (current_date)

Other than dates, deno-sqlite is working great! Thanks for creating it.

dyedgreen:
Closes #29 

Adds capability to determine current time from C wrapper/ VFS. The time is updated manually from JS every time a query is run.

This introduces some over-head (vs a poll-based approach), but is currently the only way to keep the WASM binary from having WASI dependencies (or some other import requirements).

Benchmarks seem to be not affected very much (as the update of a 64-bit float should be trivial.)

## To Do
- [x] Support binding `Date`s
- [x] Add unit tests
- [x] Add example of using `Date`s with the wrapper to `docs/examples.md`


dyedgreen:
@ricgagliardi, I believe you originally raised the issue. This is still missing some tests, but I think the problem is pretty much resolved now.

dyedgreen:
Note: Supporting dates makes it make sense to add support for directly binding `Date` objects.

Issue #34, opened 2020-03-23
Question: Why not browser compatible


timonson:
Hi @dyedgreen, I am excited to test your great project in the coming days.

While going through your documentation I have become curious about one question in particular. And I gotta say, I don't know very much about databases and wasm yet, so it might be a silly question. But I am wondering what exactly the reason for the incompatibility of deno-sqlite with the browser is. 

Considering that deno's APIs are largely browser compatible and that you chose a Virtual File System instead of persistent files, my best guess is that you can't import wasm files as ES modules. Is this the reason? 

Thank you very much!

dyedgreen:
Hi, thanks for the question!

You are right, I don’t think there is anything that prevents this library from being ported to the browser in principle (apart from the open/ safe functionality, as those require a file system.

It would however require some effort to bundle and port the module for use in the browser, and I simply haven’t had the time to do it. The main thing that would need to be done is to package the WASM binary into its own es6 module (I believe that’s what Deno is also doing internally).

This would then allow you to use SQLite directly in the browser.

Are you interested in looking into this, or would need browser compatibility for some project? 😊

timonson:
Thanks for the quick response!

I think in the first place I was just really curious but I can definitely see me using sqlite in the browser at some point in the future.

I would imagine exporting the created database as a JavaScript typed array from the browser would be a good replacement for `saving` the database.

I will try to figure out if this would be a good task for me.

Thanks again!

timonson:
Hi @dyedgreen, I really wanted to do this today but it is more complicated than I thought because browsers don't accept **top-level await** yet.
The problem really emerged when I tried to do what you said in terms of implementing the WASM binary into an ES module by converting it to **base64** first and then to JavaScript via the [new WebAssembly](https://developer.mozilla.org/en-US/docs/WebAssembly/Using_the_JavaScript_API) API.
The big obstacle is that `WebAssembly.compile()` returns a promise and it is kinda difficult to deal with promises in ES modules. I found out that `deno bundle` does exactly that: It uses `await WebAssembly.compile()` as top-level await because it works in deno.

Now we could theoretically use the dynamic import function `import()` and wait for the promise at each step. But I think I will prefer to wait until browser offer top-level await. It is already possible to use it in Chrome's [DevTools](https://v8.dev/features/top-level-await) so I guess it won't take much time until it available in the browser.

dyedgreen:
[...]

PR #42, opened 2020-05-11
Web Playground

dyedgreen:
Based on feedback from #34


dyedgreen:
I tried the `deno bundle` thing, and threw together this quick demo. Feedback welcome @timonson. (This is not an officially supported web-compatible version, but uses `deno bundle` to make a demo website, where the library can be tested, sans open/ close functions)

Issue #54, opened 2020-05-25
.get() when empty, and a .getAll() Method


tarsupin:
I'm using deno-sqlite for a webserver. So far, everything seems to be working as intended, and since this is a new project I figured I'd lend my experiences on its early development.

The largest issue comes from the fact that I'm building an SQL wrapper, which doesn't currently play nice with deno-sqlite. I need an SQL wrapper in case I switch from sqlite to an alternative at a later date. This comes with some problems, such as making the .get() methods essential to wrapping. Since I'd prefer a static entry, it also means each method has to handle the full query from db.open to db.close, and I don't have a benchmark of that cost which is concerning for large-scale use.

**.get() method**
Technically, the biggest obstruction for a wrapper right now is the uncaught TypeError (Cannot read property '_wasm' of null) that occurs from .get() on an empty select. Requires a try/catch and unpleasant handling, especially if the wrapper is trying to send a full array of results (which mine is). I believe you've already mentioned that issues with catching will be addressed in future updates, so I'm hopeful it will extend to ._get() returning an empty object when no selects are found.

**.getAll() method**
A .getAll() method would GREATLY improve support for SQL select wrappers. Without it, I'm afraid that deno-sqlite won't be usable for my circumstances, and I suspect for most wrappers. Resolving this should fix that.

tarsupin:
For an example of what I'm looking for in a wrapper:

```
	static async select( table: string, sql: string, sqlArray: Array<any> ): Promise<any[]> {
		const db = await open(SQLHandler.DBPath + table + ".db");
		const handler = db.query(sql, sqlArray);
		const results = handler.getAll();
		handler.done();
		db.close();
		return results;
	}
```

tarsupin:
I looked through the code and built a solution in rows.js (should this be in ts?) It might make sense to move the status into a shared method.

```
 getAll( limit = 10000 ) {
	let rows = [];
	for(let i = 0; i < limit; i++) {
		const row = this._get();
		if(row.length == 0) { break; }
		const status = this._db._wasm.step(this._db._id, this._id);
		switch (status) {
			case constants.Status.SqliteRow:
				// NO OP
				break;
			case constants.Status.SqliteDone:
				this.done();
				break;
			default:
				this.done();
				throw this._db._error(status);
				break;
		}
		rows.push(row);
	}
	return rows;
  }
```

tarsupin:
[...]

PR #55, opened 2020-05-26
.getAll() added, fix to ._get()

tarsupin:
Resolves https://github.com/dyedgreen/deno-sqlite/issues/54


tarsupin:
This update isn't needed, as the solutions I required can be achieved without the .next() method in select, which is the only case where the .get() method required a change.

PR #56, opened 2020-05-29
Update README to reflect second parameter for query method

Coteh:
Even if you have no arguments to fill in, you need to provide empty array otherwise you'll get the following error:
```
Expected 2 arguments, but got 1. ts(2554)
<long hash that I assume resolves to file with query method>(110, 14): An argument for 'values' was not provided.
```
Just thought I'd provide this PR to update it to make it clear for first-time viewers of this project like myself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ import { open, save } from "https://deno.land/x/sqlite/mod.ts";
 
 // Open a database
 const db = await open("test.db");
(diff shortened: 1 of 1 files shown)


dyedgreen:
Hey, thank you very much for flagging this up! The `query` method should indeed only need one parameter (as advertised in the README), so this is not a problem with the README, but rather a bug in the library.

Could you provide a minimal example of how to reproduce this, so that we can find a fix for this issue?

Coteh:
Thanks for the correction! I just created a new issue (#58) to continue this discussion, and I provided a reproducible example as well. I will close this as I misunderstood here.

PR #57, opened 2020-05-29
Start work on deno vfs

### LINKED ISSUE #49: Upgrade WASI VFS to directly interact with files
# Idea

Update the [VFS](https://github.com/dyedgreen/deno-sqlite/blob/master/build/src/vfs.c) to use imported functions, provided by a small wrapper, using the Deno file-system apis.

### Pros:
- gain strong no-corrupt databases guarantees given by SQLite (**this is a big one**)
- fast + automatic file-saving on database action
- ability to work with databases large enough to not fit into memory (**this is a big one**)
- module would be more in-line with what most people expect when using SQLite modules in other languages (see e.g. #50)

### Cons:
- won't be able to utilise direct WASM imports in Deno in the future (this is one of the original reasons against such a design)
- queries perform synchronous file-system writes (unless they use a `:memory:` database)
   - this could be mitigated by also providing an AsyncDB version (or maybe a SyncDB as the base and DB is async), which wraps the other DB and runs it in a worker, sending queries there, returning the queries etc
- nice APIs we have right now (pre-load in-memory db, grab db memory directly) will be harder to implement
- these may be slower than currently (since right now everything is in-memory, avoiding any re-direction), and it may be hard to implement the `.data()` api with the new VFS (currently it's easy, since we hold the file data in memory at a location controlled by us)
- requires breaking API changes
   - but those are mostly constraint to how the DB is opened (and with the constructor being `new DB(path = ":memory:")`, we can even preserve how in-memory DBs are opened)
   - if these changes are made we should notify our dependents

### Proposed API changes
- drop `open`
- drop `save`
- drop `DB.data()`
- change constructor `new DB(path = ":memory:")`

### List of dependents to notify of the changes (and discuss changes with)
- https://github.com/eveningkid/denodb/blob/master/deps.ts
- https://github.com/denjucks/dexecutor/tree/master/clients/libs/sqlite
- https://github.com/halvardssm/deno-nessie/blob/master/deps.ts

### LINKED ISSUE #58: TypeScript compiler flags error for query method on VS Code - ts(2554)
## Description
Originally filed as PR #56 since I mistakenly assumed that the `query` method won't accept just one argument, but this is not the case. Seems like this is an error with the typing definition for this method.

## Expected Behaviour
When I call `query` method with just the query string, there should be no TypeScript error.

## Actual Behaviour
```
Expected 2 arguments, but got 1. ts(2554)
<long hash that I assume resolves to file with query method>(110, 14): An argument for 'values' was not provided.
```
However, if you execute it with `deno run` it'll still compile and run successfully.

## Reproduction Steps
Open up Visual Studio Code in a new folder, create a TypeScript file (doesn't matter the name). Simply copy and paste the README example into it. :)
```
import { open, save } from "https://deno.land/x/sqlite/mod.ts";

// Open a database
const db = await open("test.db");
db.query("CREATE TABLE IF NOT EXISTS people (id INTEGER PRIMARY KEY AUTOINCREMENT, name TEXT)");

const names = ["Peter Parker", "Clark Kent", "Bruce Wayne"];

// Run a simple query
for (const name of names)
  db.query("INSERT INTO people (name) VALUES (?)", [name]);

// Print out data in table
for (const [name] of db.query("SELECT name FROM people"))
  console.log(name);

// Save and close connection
await save(db);
db.close();
```
Also have the [Deno VS Code extension](https://marketplace.visualstudio.com/items?itemName=denoland.vscode-deno) installed and ensure it's enabled on your project by putting this in either workspace or user `settings.json`:
```
{
    "deno.enable": true
}
```

dyedgreen:
Update VFS
Closes #58 
Closes #49


dyedgreen:
Closes #58

PR #60, opened 2020-06-01
Add lastInsertRowId to DB instance

rahmanfadhil:
I'm working on an open-source ORM for Deno called [Cotton](https://github.com/rahmanfadhil/cotton). It relies on this package, and I need to have access to the last inserted row id whenever I insert a new record to the database.

It basically exposes the `sqlite3_last_insert_rowid` function, and the value is now accessible in the `lastInsertRowId` property.

Let me know if I need to do anything.


build/src/wrapper.c
```diff
@@ -188,3 +188,7 @@ const char* EXPORT(column_origin_name) (sqlite3_stmt* stmt, int col) {
 const char* EXPORT(column_table_name) (sqlite3_stmt* stmt, int col) {
   return sqlite3_column_table_name(stmt, col);
 }
+
+double EXPORT(last_insert_rowid) () {
+  return sqlite3_last_insert_rowid(database);
```
dyedgreen, line 193: Please explicitly cast to double:
```suggestion
  return (double)sqlite3_last_insert_rowid(database);
```

src/db.ts
```diff
@@ -230,6 +230,13 @@ export class DB {
     this._open = false;
   }
 
+  /**
+   * Get last inserted row id
```
dyedgreen, line 234: Please format the comment such that it will be included in the generated documentation, and also be specific about what SQLite function this corresponds to:
```suggestion
   * DB.lastInsertRowId
   *
   * Get last inserted row id. This corresponds to
   * the SQLite function `sqlite3_last_insert_rowid`.
```

src/db.ts
```diff
@@ -230,6 +230,13 @@ export class DB {
...
+   */
+  get lastInsertRowId(): number {
+    return this._wasm.last_insert_rowid();
```
dyedgreen, line 237: I'm not sure on this, but do you think we should throw an error if a user attempts to obtain the last insert row id from a closed DB?

rahmanfadhil: I don't think we have to do that. If the user tries to access it when the database is closed, the value will be reset to 0 again.

src/db.ts
rahmanfadhil: Fixed

build/src/wrapper.c
rahmanfadhil: Fixed

test.ts
```diff
@@ -579,3 +579,23 @@ Deno.test("dateTimeIsCorrect", function () {
...
+
+  // By default, lastInsertRowId must be 0
+  assertEquals(db.lastInsertRowId, 0);
+
+  // Create table and insert value
+  db.query(`CREATE TABLE users (id INTEGER PRIMARY KEY, name VARCHAR(255))`);
+  db.query("INSERT INTO users (name) VALUES ('John Doe')");
+
+  // Now, the last inserted row id must be 1
+  assertEquals(db.lastInsertRowId, 1);
```
dyedgreen, line 594: I'd be happier with getting the correct value from an SQL statement, so to do something like this:

```js
const insertRowIds = [];
for (let i = 0; i < 10; i ++) {
  db.query("INSERT INTO users (name) VALUES ('John Doe')");
  insertRowIds.push(db.lastInsertRowId);
}

assertEquals(insertRowIds, [...db.query("SELECT rowid FROM users")].map(([i]) => i));
```

rahmanfadhil: Done!

PR #74, opened 2020-07-14
Added toObjects to Rows

### LINKED ISSUE #71: Provide example for getting array of objects from SELECT
Hi! I am trying to use this library for a side project, and I am having some problems using a select query and map column names to the rows so that the result is an array of objects. Could you provide an example for this? 😊

halvardssm:
fixes #71 

@dyedgreen Here is a PR, let me know what you think 😄 The one thing I am not too sure about is the error catching on lines 153-160, would there be a neater way of dealing with this or should I just leave it as such?


dyedgreen:
Thank you for the PR! Please bare with me as, I try to have a look through it some time tonight or tomorrow 😅

src/error.ts
```diff
@@ -1,5 +1,8 @@
 import { Status } from "./constants.ts";
 
+export const ERROR_TRANSACTION_FINALIZED =
```
dyedgreen, line 3: I'm not sure if I like this or not, but generally: if we want to move all error strings into one place, we should do it in a dedicated pull (there are some in other places as well).

Generally I might prefer to not do this though, so we can keep the error strings close to where the error originates. (They don't need to/ shouldn't be accessible for e.g. programmatically checking what the error is, since they are meant for human consumption and not part of the stable API.)

src/rows.ts
```diff
@@ -1,6 +1,6 @@
 import { getStr } from "./wasm.ts";
 import { Status, Values, Types } from "./constants.ts";
-import SqliteError from "./error.ts";
+import SqliteError, { ERROR_TRANSACTION_FINALIZED } from "./error.ts";
```
dyedgreen, line 3: See above comment :)

src/rows.ts
```diff
@@ -102,7 +102,7 @@ export class Rows {
   columns(): ColumnName[] {
     if (this._done) {
       throw new SqliteError(
-        "Unable to retrieve column names as transaction is finalized.",
+        ERROR_TRANSACTION_FINALIZED,
```
dyedgreen, line 105: See above comment :)

src/rows.ts
```diff
@@ -126,6 +126,41 @@ export class Rows {
...
+  /**
+   * Rows.toObjects
+   * 
+   * Call this if you need to ouput the rows as objects.
+   * 
+   * Will return an empty array if there are no entries in the table.
+   * 
+   *     const rows = db.query("SELECT name FROM users;").toObjects();
+   */
+  toObjects<T extends any = Record<string, any>>(): T[] {
```
dyedgreen, line 138: While I like the signature, I don't like the implementation. I'd prefer to return an iterator here, so that there is no big up-front read necessary. (If a user wants an array, they can use the spread operator: `[...db.query('SELECT * FROM people').asObjects()]`)

I was thinking of having a new class (e.g. `RowObjects`?, feel free to come up with a better name 😅 ), which stores the columns out put and basically consumes the underlying `Rows` iterator, transforming each row to an object as it passes through.

For that implementation I think I would also like to use `asObjects`, rather `toObjects`.

test.ts
```diff
@@ -627,3 +627,49 @@ Deno.test("lastInsertedId", function () {
...
   assertEquals(db.lastInsertRowId, 0);
 });
+
+Deno.test("outputToObjectArray", function () {
+  const db = new DB();
+
+  const expectedName = "John Doe";
+
+  // Create table and insert value
+  db.query(`CREATE TABLE users (id INTEGER PRIMARY KEY, name VARCHAR(255))`);
```
dyedgreen, line 637: Use `TEXT` instead of `VARCHAR` (see https://www.sqlite.org/datatype3.html#affinity_name_examples)

test.ts
```diff
@@ -627,3 +627,49 @@ Deno.test("lastInsertedId", function () {
...
+    assert(row.name === expectedName, "Name is incorrect");
+    assert(typeof row.id === "number", "ID is incorrect");
+  }
+});
+
+Deno.test("outputToObjectArrayEmpty", function () {
+  const db = new DB();
+
+  // Create table and insert value
+  db.query(`CREATE TABLE users (id INTEGER PRIMARY KEY, name VARCHAR(255))`);
```
dyedgreen, line 666: See above.

dyedgreen:
Thank you for the PR! (And sorry for taking so long to get to reviewing it, I'm quite busy with work at the moment 😅) Let me know if there is anything in the review that isn't clear 😃

src/rows.ts
dyedgreen: I was just thinking about the try catch: You should be able to check `this === Empty` instead 😃

src/error.ts
halvardssm: No problem, I changed it back :)

src/rows.ts
halvardssm: same as above :)

src/rows.ts
halvardssm: same as above

test.ts
halvardssm: 👍

test.ts
halvardssm: 👍

src/rows.ts
halvardssm: Fair enough. The only thing I am not sure if I am too fond of is to create a new class to handle this. Wouldn't it be better to add an option argument to `query()`? If you strongly prefer to create a new class, I can do that, but I think it would keep the codebase cleaner if we add an argument to `query()`, add a private property to the class called `_columns` (or something) so that we only have to call `columns()` once,  and alter `next()` to facilitate this change. Let me know what you think! 😄

halvardssm:
> Thank you for the PR! (And sorry for taking so long to get to reviewing it, I'm quite busy with work at the moment 😅) Let me know if there is anything in the review that isn't clear 😃

No worries! I know the feeling of juggling a dousin of things at the same time 👍

src/rows.ts
halvardssm: I just pushed a commit where you can see what I mean above. If you prefer to use a class I can revert the commit

halvardssm: I see your point, and I agree. I will change it as per the first option, but I might need some feedback in regards to the iterator (as this will be the first time I actually write one) 😄

dyedgreen: No worries, that’s what code reviews are for 😀

halvardssm: So I implemented RowObjects 🎉 Let me know if there are any things to improve!

src/db.ts
```diff
@@ -2,7 +2,7 @@ import instantiate from "../build/sqlite.js";
 import { getStr, setStr, setArr } from "./wasm.ts";
 import { Status, Values } from "./constants.ts";
 import SqliteError from "./error.ts";
-import { Rows, Empty } from "./rows.ts";
+import { Rows, Empty, RowsOptions } from "./rows.ts";
```
dyedgreen, line 5: I think you still need to remove this 😅

src/db.ts
```diff
@@ -15,6 +15,8 @@ type QueryParam =
   | Date
   | Uint8Array;
 
+type QueryOptions = RowsOptions;
```
dyedgreen, line 18: See above

src/db.ts
```diff
@@ -110,7 +112,11 @@ export class DB {
    * iterated over or discarded by calling
    * `.return()` or closing the iterator.
    */
-  query(sql: string, values?: object | QueryParam[]): Rows {
+  query(
+    sql: string,
+    values?: object | QueryParam[],
+    options?: QueryOptions,
```
dyedgreen, line 118: See above

src/db.ts
```diff
@@ -208,7 +214,7 @@ export class DB {
         return Empty;
         break;
       case Status.SqliteRow:
-        const transaction = new Rows(this, stmt);
+        const transaction = new Rows(this, stmt, options);
```
dyedgreen, line 217: See above

src/rows.ts
```diff
@@ -1,17 +1,24 @@
...
+import { RowObjects } from "./row_objects.ts";
 
-interface ColumnName {
+export interface ColumnName {
   name: string;
   originName: string;
   tableName: string;
 }
 
+export interface RowsOptions {
```
dyedgreen, line 12: See above

src/rows.ts
```diff
@@ -1,17 +1,24 @@
...
+  asObjects: boolean;
+}
+
 export class Rows {
   private _db: any;
   private _stmt: number;
   private _done: boolean;
+  private _options?: RowsOptions;
```
dyedgreen, line 20: See above

src/rows.ts
```diff
@@ -1,17 +1,24 @@
...
+  private _columns?: ColumnName[];
```
dyedgreen, line 21: I think this should live in the `RowObjects` object 😃

src/rows.ts
```diff
@@ -24,14 +31,19 @@ export class Rows {
    * and the only correct way to obtain a `Rows`
    * object is by making a database query.
    */
-  constructor(db: any, stmt: number) {
+  constructor(db: any, stmt: number, options?: RowsOptions) {
```
dyedgreen, line 34: see above

src/rows.ts
```diff
@@ -24,14 +31,19 @@ export class Rows {
...
     this._db = db;
     this._stmt = stmt;
+    this._options = options;
     this._done = false;
 
     if (!this._db) {
       this._done = true;
     }
+
+    if (this._options?.asObjects) {
```
dyedgreen, line 44: See above

src/rows.ts
```diff
@@ -83,7 +95,17 @@ export class Rows {
         throw this._db._error(status);
         break;
     }
-    return { value: row, done: false };
+
+    if (this._options?.asObjects) {
```
dyedgreen, line 99: Yeah, I think you might have left this in from the previous commit. I think this should live exclusively inside the `RowObjects`

src/rows.ts
```diff
@@ -126,6 +148,17 @@ export class Rows {
...
   }
 
+  /**
+   * Rows.asObjects
+   * 
+   * Call this if you need to ouput the rows as objects.
+   * 
+   *     const rows = [...db.query("SELECT name FROM users;").asObjects()];
+   */
+  asObjects<T extends any = Record<string, any>>(): RowObjects<T> {
```
dyedgreen, line 158: I think the type annotation should live in `RowObjects` and shouldn't be generic (as we always return the same: an iterator over `Record`s)

test.ts
```diff
@@ -740,3 +740,49 @@ Deno.test("changes", function () {
...
+
+  // Insert data to table
+  for (let i = 0; i < 2; i++) {
+    db.query("INSERT INTO users (name) VALUES ('John Doe')");
+  }
+
+  const res = [...db.query("SELECT * FROM users").asObjects()];
+
+  assert(
+    typeof res === "object" && res.length === 2,
```
dyedgreen, line 760: Not sure about the typeof, since that's basically testing your tests code

test.ts
```diff
@@ -740,3 +740,49 @@ Deno.test("changes", function () {
...
+  const db = new DB();
+
+  // Create table and insert value
+  db.query(`CREATE TABLE users (id INTEGER PRIMARY KEY, name TEXT)`);
+
+  // All collected row ids must be the same as in the database
+  const res = [...db.query("SELECT * FROM users").asObjects()];
+
+  assert(
+    typeof res === "object" && res.length === 0,
```
dyedgreen, line 785: See above comment on typeof

src/row_objects.ts
```diff
@@ -0,0 +1,66 @@
+import { Rows, Empty, ColumnName } from "./rows.ts";
+
+export class RowObjects<T extends any = Record<string, any>> {
```
dyedgreen, line 3: I don't think this should be generic, since the type never changes 😅

src/row_objects.ts
```diff
@@ -0,0 +1,66 @@
...
+  private _rows: Rows;
+  private _columns?: ColumnName[];
+  private _done = false;
```
dyedgreen, line 6: The `_done` state should be managed by the underlying `Rows` iterator 😃

src/row_objects.ts
```diff
@@ -0,0 +1,66 @@
...
+    return this._rows.return();
+  }
+
+  /**
+   * RowObjects.next
+   *
+   * Implements the iterator protocol.
+   */
+  next(): IteratorResult<T> {
+    if (this._done) return this._rows.return();
```
dyedgreen, line 48: You can do
```
const { value, done } = this._rows.next();
if (done) {
   return { value: undefined, done: true };
}
```

src/row_objects.ts
halvardssm: Actually, by having this generic the user can use their own types, e.g. `asObjects<CustomObject>`, I could maybe change the signature though, what do you think? 🤔

src/db.ts
halvardssm: 👍

halvardssm: 👍

src/row_objects.ts
halvardssm: You are so right! I didn't think about that 😅

src/row_objects.ts
dyedgreen: Ohhh, I see 👀 Yeah, you're right, I didn't think of that, let's keep the type parameter 😅

tksilicon:
I got this error which happens when the database is empty _SqliteError: Unable to retrieve column names as transaction is finalized_ Seems @halvardssm solved it with this pull request and is waiting for @dyedgreen to merge. I am using this lib for [DSO](https://deno.land/x/dso).

halvardssm:
> lgtm 😍 
> 
> 
> 
> Let me merge this tomorrow, when I'm less sleepy 😴 
> 
> 
> 
> (and thanks for bearing with me through the slow reviews)


No worries, I haven't had too much time lately myself (I guess this is the curse of open source) 😄 

Thanks for the pointers regarding implementing the iterator! 💪🏻

Issue #80, opened 2020-07-24
Transactions


tksilicon:
How is transaction implemented? In this [mysql](https://deno.land/x/mysql) connection object is expose to handle transactions. How does it work in this library? This library is being plugged into [DSO](https://deno.land/x/dso).

dyedgreen:
You can use the [standard SQLite way](https://www.sqlite.org/lang_transaction.html) of doing it:
```JS
db.query('begin');
// run some queries ...
db.query('commit');
```

dyedgreen:
(Btw, feel free to reach out if you need any help or encounter any issues while trying to include the library, and let me know once you're done, so I can include DSO in the README 😄  )

tksilicon:
This is an ORM and will be using models to run transactions not SQL. There should be way for the client to implement the begin and commit and expose and then expose a connection that can be passed to the model. We can't have users running SQL when it's an ORM.

dyedgreen:
Yeah, the calls should probably be made in the ORM layer. So your ORM should do the following:
```JS
transaction(async () => { // at this point the ORM should run db.query('begin')
  model.query1(); // run generated SQL ...
  model.query2();
  model.query3();
  return boolValue; // at this point conditionally run db.query(returnedTrue ? 'commit' : 'rollback')
});
```

dyedgreen:
```JS
// example implementation of synchronous transaction:
class Client {
  construct(db) {
    this._db = db;
  }

  transaction(callback) {
    this._db.query('begin');
    try {
      const commit = callback();
      this._db.query(commit ? 'commit' : 'rollback');
    } catch {
      this._db.query('rollback');
    }
  }

  stubQuery() {
    // pretend this runs a query
  }
}

// use it!
const conn = new Client(myDb);

conn.transaction(() => {
  conn.stubQuery();
  return true;
});
```

tksilicon:
> Yeah, the calls should probably be made in the ORM layer. So your ORM should do the following:
> 
> ```js
> transaction(async () => { // at this point the ORM should run db.query('begin')
>   model.query1(); // run generated SQL ...
>   model.query2();
>   model.query3();
>   return boolValue; // at this point conditionally run db.query(returnedTrue ? 'commit' : 'rollback')
> });
> ```

transaction(async () => { // at this point the ORM should run db.query('begin')
  model.insert(); // run generated SQL ...
  model.update();
  model.insert();
[...]

PR #81, opened 2020-07-30
Add transaction properties to DB client

tksilicon:
I opened this issue #80 which requires a mechanism to run transactions in [DSO ORM](https://github.com/manyuanrong/dso) . What I did was to use a modified local class: .

Here are examples of its usage.

```js
clientTestSQLITE(async function testTransactionSuccess() {
  let topicId: number | undefined;
  let userId: number | undefined;
  const result = await dso.transactionSqlite<boolean>(async (trans) => {
    const userModel = trans.getModel(UserModel);
    const topicModel = trans.getModel(TopicModel);
    userId = await userModel.insert(
      { nickName: "foo", password: "bar" },
    );
    topicId = await topicModel.insert({ title: "zoo", userId });
    let user = await userModel.findById(userId!);
    assert(!!user);
    return true;
  });
  const user = await userModel.findById(userId!);
  const topic = await userModel.findById(topicId!);
  assertEquals(result, true);
  assert(!!topic);
  assert(!!user);
});

clientTestSQLITE(async function testTransactionFail() {
  let userId: number | undefined;
  let topicId: number | undefined;
  await assertThrowsAsync(async () => {
    await dso.transactionSqlite<boolean>(async (trans) => {
      const userModel = trans.getModel(UserModel);
      const topicModel = trans.getModel(TopicModel);
      userId = await userModel.insert({ nickName: "foo", password: "bar" });
      topicId = await topicModel.insert({ title: "zoo", userId });
      let user = await userModel.findById(userId!);
      assert(!!user);
      await userModel.query(new Query().table("notexists").select("*"));
      return true;
    });
  });
```

I had to [localise the client](https://gist.github.com/tksilicon/1b58b8f29333c386c040a83c20e4164d), [deps](https://gist.github.com/tksilicon/d984935064acb20b74dddaec7c253ee5) and other boilerplate code. It would be good to have this capability in-built in the client, straight from this library.

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -3,6 +3,12 @@ import { getStr, setStr, setArr } from "./wasm.ts";
 import { Status, Values } from "./constants.ts";
 import SqliteError from "./error.ts";
 import { Rows, Empty } from "./rows.ts";
+import { log } from "https://deno.land/x/mysql/src/logger.ts";
+
+/** Transaction processor */
+export interface TransactionProcessor<T> {
+  (connection: DB): Promise<T>;
+}
 
 // Possible parameters to be bound to a query
 type QueryParam =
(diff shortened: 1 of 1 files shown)


dyedgreen:
Thank you so much for your contribution @tksilicon, I really appreciate your effort in writing and submitting this PR! However, I feel like the proposed changes do not go well with the overall [technical direction](https://dyedgreen.github.io/deno-sqlite/#/contributing?id=technical-direction) of the library.

Generally, I don't want to wrap SQL commands or provide higher-level abstractions that go beyond running SQL queries and returning results from them.

I am excited that you want to use this library for DSO ORM and would be very happy to help you with writing a compatible client layer for your library that interfaces with this libraries lower-level APIs. However, I strongly feel that such an interface should be part of the ORM, since each ORM will have unique opinions about their architecture and interface requirements.

tksilicon:
Where does the leave DSO ORM? Means we have to make use of localised class because the ORM needs a way to run transactions at a high level using callbacks with the SQL abstracted.

dyedgreen:
I would recommend to write a client that uses an underlying open DB connection. If you can link me to the interface you have to implement, I'd be happy to provide you with an example of how to go about implementing something like that. But generally the following would be the pattern (note that I don't know what interface you have to implement exactly, but the general idea is to write a wrapping client around the database connection, see also https://github.com/denolib/typeorm/tree/master/src/driver/sqlite for a further example of how to accomplish this):

```js
import {DB} from 'https://deno.land/x/sqlite/mod.ts';

class Client {

  construct(path) {
    this.db = new DB(path);
  }

  async query(sql) {
    return [...this.db.query(sql)];
  }

  async transaction(fn) {
    this.db.query('begin');
    try {
      await fn(this);
      this.db.query('commit');
    } catch () {
      this.db.query('rollback');
    }
  }

}

// use like
const cli = new Client(':memory:');
await cli.transaction(async (cli) => {
  await cli.query('...');
});
```

tksilicon:
Will try this approach. Hopefully it fits into our existing design.

PR #82, opened 2020-08-30
Create devcontainer.json

dyedgreen:


dyedgreen:
TODO: Should use a docker image to install deno and correct WASI + compiler toolchains + make

Issue #20, opened 2020-02-11
Add json1 support


jolonf:
Include the sqlite json1 extension:

https://www.sqlite.org/json1.html

dyedgreen:
Thank you for opening an issue @jolonf!

I'm not sure the added build complexity/ performance impact of the extension is worth the benefit. Especially since JavaScript/ TypeScript has very strong JSON support already (as opposed to say C, where these utilities are much more helpful).

Could you elaborate on your use case for the extension?

jolonf:
It looks like deno-sqlite is only memory-based at the moment? In that case there probably isn't much of an advantage to having the json1 extension.

For file-based DBs the json1 extension allows json paths to be indexed for efficient retrieval.

hsjoberg:
@jolonf `deno-sqlite` is not only memory based, you need to manually call `save` function with the db as argument.

dyedgreen:
@jolonf, @hsjoberg Yeah, that is correct. Although internally the database connections are file-based, these files are completely represented and held inside the WASM memory and not saved to disc automatically. (This is probably the most significant difference from standard SQLite you'd have in C).

From what I can tell, most of the functions in the json1 extension are not really necessary when the data comes from JS, as the same manipulations can be done in JS-land easily. Although the ability to index and select on the JSON contents might be helpful to some users.

But I feel that overall, the benefit probably does not justify the added maintenance burden and the performance/ size impact on consumers who don't use the extension.

I think a good way forward here might be to have more extensive documentation and guidance on how to make customised builds, so that advanced users can alter the loaded extensions to their specific needs (see also issue #25).

jeremyBanks:
I would like to encourage you to reconsider. 🙂

I am currently using Node with the NPM `sqlite3` package for a small project, because it supports the JSON functions out-of-the-box. I would have needed to do a custom build to enable them for Python or Ruby's primary SQLite libraries, and that seemed daunting.

Thanks to your work on this repository, and Deno's and SQLite's simplicity, I was pleased to find that making a custom build with JSON1 enabled was very easy, just a matter of changing a few lines: b0214ce46a1d2a9d143b4c016c7df4dfaede10e6. 🙏 My needs are satisfied, but I think there are a lot of potential users who would appreciating having it enabled by default, as [it is in the competing NPM library](https://www.npmjs.com/package/sqlite3#features).

[...]

PR #88, opened 2020-10-28
Enable JSON1 extension for SQLite

jeremyBanks:
This was previously declined in #20, but I would like to encourage you to reconsider. 🙂

I am currently using Node with the NPM `sqlite3` package for a small project, because it supports the JSON functions out-of-the-box. (I would have needed to do a custom build to enable them for Python or Ruby's primary SQLite libraries, and that seemed daunting.) I would have liked to have chosen Deno instead, but this library didn't support them.

I use SQLite as an in-memory database to provide indexing and data validation/integrity guarantees. Like many JavaScript developers, the data I work with often comes in the form of loosely-structured JSON objects, for better or worse. I find it extremely valuable to be declare constraints and indices against JSON properties, gradually-typing my data in a TypeScripty fashion while having SQLite confirm my assumptions against the real data. It's nice to be able to have all of this validation logic expressed together in SQL, instead of needing to have one set of logic (or transformations) in JavaScript for the JSON objects (maybe using io-js or runtypes), and a separate set of validation logic in SQL.

Fortunately, it seems like enabling this extension statically doesn't require compiling any of the dynamic extension support, so the impact on build complexity and size is small: it just requires adding one build flag, and the compiled WASM is only 3% larger. Given the prevalence of JSON data in JavaScript, I suggest this capability may be worth that cost.

```
sqlite.wasm:    621987B ➜ 642364B (+3.2%)
sqlite.js:      829955B ➜ 857127B (+3.2%)
gzip sqlite.js: 339540B ➜ 349630B (+3.0%)
```

Thank you for your work on this library.

(The random style changes in here were made by `deno fmt`, in order to satisfy the CI.)


dyedgreen:
Hi @jeremyBanks, thank you so much for submitting this pr, and apologies for taking so long to get back to you.

I'm currently quite busy, with exams coming up, but I just had some time to fix one of the current issues and wanted to update you on this issue:

Regarding the added binary size, I'm not too concerned with this, since the change is quite small and this is a server side library. My main issue was with added complexity / maintenance burden, but looking at the PR, this seems a matter of adding a simple build flag, so that should actually also be fine.

I'll give your change a proper code-review later, but from having a quick look, here is some things I'd like to change before merging this PR 😅

- You updated the dev container, this should probably be a separate PR
- You're changing the compiled binary, which I'd prefer to automate

The build stuff is not your fault, but in short I would want to add this as a CI step before accepting the PR, so that users of the library don't have to trust that the binary is not malicious in some way (which the CI let's them verify). This is mostly fine since Deno is sand-boxed, but adding a little extra verify-ability never hurts. Plus it makes contributing C changes easier, since you can't accidentally commit a debug build to master. (To be clear, this is also an issue with the binary built locally by myself, it's just now surfaced again when looking through this PR 😄)

So in short: Please bear with me to upgrade the CI (or feel free to submit a PR that does that 😉)

dyedgreen (changes requested):
See previous comment 😅

jeremyBanks:
@dyedgreen No worries, I appreciate you taking the time to update me. You raise good points; I do feel a little weird committing a binary I compiled locally. 🙂 When I have some time I'll update the PR, and possibly take a stab at the CI.

test.ts
```diff
@@ -780,6 +780,40 @@ Deno.test("outputToObjectArrayEmpty", function () {
   );
 });
 
+Deno.test("jsonFunctions", function () {
+  const db = new DB();
+
+  // The JSON1 functions should exist and we should be able to call them without unexpected errors
+  db.query(`SELECT json('["this is json"]')`);
+  db.query(`SELECT json('{"this is": "json"}')`);
+  db.query(`SELECT json('2')`);
```
dyedgreen, line 789: I think any one of these should be fine. We don't really need to unit test SQLite itself 😄

dyedgreen (commented):
This looks good to me, but lets wait for #95 😅

test.ts
jeremyBanks: done

dyedgreen:
This looks good to me! Can you rebase on master, so I can merge it? 😃

dyedgreen (approved):

jeremyBanks:
Feel free to squash if you want, I know these PRs have become pretty messy, up to you.

dyedgreen:
Actually, can you build the WASM again? (I think you deleted the files, and it's causing the tests to fail 😅). Then I'll squash and merge 👍

dyedgreen:
Alright, if you rebase on master it should go through now? 😃

jeremyBanks:
I think that the WASM from `master` would still fail because of our new check for JSON functions.

jeremyBanks:
@dyedgreen I think we're good. :) Your approach should prevent the need for re-building before merging to master.

This PR's workflow changes make it possible to do so manually, as I have done here at https://github.com/jeremyBanks/deno-sqlite/actions/runs/496785163, but feel free to revert that if you think it's out of scope or unnecessary.

dyedgreen:
> This PR's workflow changes make it possible to do so manually, as I have done here at https://github.com/jeremyBanks/deno-sqlite/actions/runs/496785163, but feel free to revert the workflow changes here if you think it's out of scope or unnecessary. (You should have write access to this PR branch, if you feel like doing that.)

I reverted the changes to the workflow, since running it manually should not be necessary. (The test runner should work without any tampering 😅 )

dyedgreen:
Thanks again for the PR @jeremyBanks! 😍

PR #91, opened 2020-11-01
Fix binding of +Infinity, -Infinity, -0, and extremely large number values

### LINKED ISSUE #89: ±Infinity are replaced with -9223372036854775808
👋 Hello! Thank you for the very useful library. I have a bug report.

SQLite's `REAL` values are IEEE floating point numbers, and although they do not support `NaN`s, they do support infinities. An easy way to produce them is with overflowing float literals:

```ts
#!/usr/bin/env -S deno run
import { DB } from "https://deno.land/x/sqlite@v2.3.0/mod.ts";
const db = new DB();

for (const sqlValue of ['1e34567', '-1e34567']) {
  const [[result]] = db.query(`SELECT ${sqlValue}`);
  console.log(sqlValue, '->', result);
}
```

```
1e34567 -> Infinity
-1e34567 -> -Infinity
```

This library returns ±Infinity values from SQLite correctly. However, when passing them to SQLite, they are converted to the value `-9223372036854775808`.

```ts
for (const jsValue of [1e34567, -1e34567]) {
  const [[result]] = db.query("SELECT ?", [jsValue]);
  console.log(jsValue, '->', result);
}
```

```
Infinity -> -9223372036854775808n
-Infinity -> -9223372036854775808n
```

jeremyBanks:
Closes #89 

Fixes and adds tests the incorrect binding of `+Infinity`, `-Infinity`, `-0`, and extremely-large `number` values, by only binding to `INTEGER` for values which satisfy ` (Number.isSafeInteger(value) && !Object.is(value, -0))`, instead of `(Math.floor(value) === value)`.

In order to be preserve `-0`, we have to bind it as a `REAL` instead of an `INTEGER` like `+0`. As far as I can tell, this should be harmless.

A change in behaviour to call out: integer number values more massive than `MAX_SAFE_INTEGER` which could fit into an `INTEGER` will now be stored as `REAL` instead of `INTEGER`. I could go either way on this, but settled with this because you're usually not supposed to be treating values larger than `MAX_SAFE_INTEGER` as integers, even if they happen to fall on integer values.

If you want to maintain them as `INTEGER`, I think you could change the condition to something like `(Number.isFinite(value) && !Object.is(value, -0) && Math.floor(value) === value && -9_223_372_036_854_775_808 <= value && value <= 9_223_372_036_854_775_807)` instead.

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -157,7 +157,7 @@ export class DB {
           value = value ? 1 : 0;
         // fall through
         case "number":
-          if (Math.floor(value) === value) {
+          if (Number.isSafeInteger(value) && !Object.is(value, -0)) {
             status = this._wasm.bind_int(stmt, i + 1, value);
           } else {
             status = this._wasm.bind_double(stmt, i + 1, value);
--- a/src/db.ts
+++ b/src/db.ts
@@ -157,7 +157,7 @@ export class DB {
           value = value ? 1 : 0;
         // fall through
         case "number":
-          if (Number.isSafeInteger(value) && !Object.is(value, -0)) {
+          if (Number.isSafeInteger(value)) {
             status = this._wasm.bind_int(stmt, i + 1, value);
           } else {
             status = this._wasm.bind_double(stmt, i + 1, value);
--- a/test.ts
+++ b/test.ts
@@ -786,3 +786,63 @@ Deno.test("outputToObjectArrayEmpty", function () {
     "Result is not an array or does not have the correct length",
   );
 });
+
+Deno.test("infinitiesAndInfinitesimals", function () {
+  const db = new DB();
+
+  // return
+
+  const [[selectedPositiveInfinity]] = db.query("SELECT +8e8888");
+  assertEquals(
+    selectedPositiveInfinity,
+    +Infinity,
+    "Failed to return correct value for +Infinity.",
+  );
+
+  const [[selectedNegativeInfinity]] = db.query("SELECT -8e8888");
+  assertEquals(
+    selectedNegativeInfinity,
+    -Infinity,
+    "Failed to return correct value for -Infinity.",
+  );
+
+  const [[selectedPositiveZero]] = db.query("SELECT +1/8e8888");
+  assert(
+    Object.is(selectedPositiveZero, +0),
+    "Failed to return correct value for +0.",
+  );
+
+  const [[selectedNegativeZero]] = db.query("SELECT -1/8e8888");
+  assert(
+    Object.is(selectedNegativeZero, -0),
+    "Failed to return correct value for -0.",
+  );
+
+  // bind
+
+  const [[boundPositiveInfinity]] = db.query("SELECT ?", [+Infinity]);
+  assertEquals(
+    boundPositiveInfinity,
+    Infinity,
+    "Failed to bind correct value for +Infinity.",
+  );
+
+  const [[boundNegativeInfinity]] = db.query("SELECT ?", [-Infinity]);
+  assertEquals(
+    boundNegativeInfinity,
+    -Infinity,
+    "Failed to bind correct value for -Infinity.",
+  );
+
+  const [[boundPositiveZero]] = db.query("SELECT ?", [+0]);
+  assert(
+    Object.is(boundPositiveZero, 0),
+    "Failed to bind correct value for +0.",
+  );
+
+  const [[boundNegativeZero]] = db.query("SELECT ?", [-0]);
+  assert(
+    Object.is(boundNegativeZero, -0),
+    "Failed to bind correct value for -0.",
+  );
+});
--- a/test.ts
+++ b/test.ts
@@ -787,7 +787,7 @@ Deno.test("outputToObjectArrayEmpty", function () {
   );
 });
 
-Deno.test("infinitiesAndInfinitesimals", function () {
+Deno.test("veryLargeAndVerySmall", function () {
   const db = new DB();
 
   // return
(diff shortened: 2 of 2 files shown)


src/db.ts
```diff
@@ -157,7 +157,7 @@ export class DB {
           value = value ? 1 : 0;
         // fall through
         case "number":
-          if (Math.floor(value) === value) {
+          if (Number.isSafeInteger(value) && !Object.is(value, -0)) {
```
dyedgreen, line 160: Can we remove the `Object.is(value, -0)` check?

You are right, that it will preserve the IEEE `-0` through a `SELECT ?` type query. But if you first store the values into a database table, and then retrieve them, the `-0` becomes a `+0` again (SQLite will be doing some serialisation, which does not preserve the difference).

Since the `-0` is not preserved through storing to a table, I don't think it makes sense to try preserving it through a `SELECT ?`, especially since you might do some arithmetic on integers and get a `-0` 'integer', which you actually want to keep as an integer 😅 

(Example of how `-0` is not persisted through table:)
```JavaScript
import { DB } from "./mod.ts";

const db = new DB();
db.query("CREATE TABLE numbers (id INTEGER PRIMARY KEY AUTOINCREMENT, num REAL)");

const val = -0;
db.query("INSERT INTO numbers (num) VALUES (?)", [val]);
const [[outVal]] = db.query("SELECT num FROM numbers");

console.log(`In: ${val}; Out: ${outVal}`);
console.log(`= ${val === outVal} : is ${Object.is(val, outVal)}`);
```

test.ts
```diff
@@ -786,3 +786,91 @@ Deno.test("outputToObjectArrayEmpty", function () {
     "Result is not an array or does not have the correct length",
   );
 });
+
+Deno.test("veryLargeAndVerySmall", function () {
+  const db = new DB();
+
+  // return
```
dyedgreen, line 793: Can we remove all of the `// return` section? They effectively only test SQLite code, which should be very well tested by itself 😅

test.ts
```diff
@@ -786,3 +786,91 @@ Deno.test("outputToObjectArrayEmpty", function () {
...
+  );
+
+  const [[selectedNegativeMassive]] = db.query("SELECT -20e20");
+  assertEquals(
+    selectedNegativeMassive,
+    -20e20,
+    "Failed to return correct value for -20e20.",
+  );
+
+  // bind
```
dyedgreen, line 835: The comment is not necessary after removing the preceding section 😅

test.ts
```diff
@@ -786,3 +786,91 @@ Deno.test("outputToObjectArrayEmpty", function () {
...
+
+  const [[boundPositiveInfinity]] = db.query("SELECT ?", [+Infinity]);
```
dyedgreen, line 837: Can we test these against a proper table. E.g.

```JavaScript
db.query("CREATE TABLE numbers (id INTEGER PRIMARY KEY AUTO INCREMENT, number REAL)");

db.query("INSERT INTO numbers (number) VALUES (?)", [+Infinity]);
db.query("INSERT INTO numbers (number) VALUES (?)", [-Infinity]);
// ...

const [[positiveInfinity], [negativeInfinity] /* ... */] = db.query("SELECT number FROM numbers");

assertEquals(positiveInfinity, +Infinity);
// ...
```

test.ts
```diff
@@ -786,3 +786,91 @@ Deno.test("outputToObjectArrayEmpty", function () {
...
+  );
+
+  const [[boundNegativeInfinity]] = db.query("SELECT ?", [-Infinity]);
+  assertEquals(
+    boundNegativeInfinity,
+    -Infinity,
+    "Failed to bind correct value for -Infinity.",
+  );
+
+  const [[boundPositiveZero]] = db.query("SELECT ?", [+0]);
```
dyedgreen, line 851: We won't need these tests, since the `-0` is not preserved by tables anyways. 😅

test.ts
```diff
@@ -786,3 +786,91 @@ Deno.test("outputToObjectArrayEmpty", function () {
...
+    "Failed to bind correct value for +0.",
+  );
+
+  const [[boundNegativeZero]] = db.query("SELECT ?", [-0]);
+  assert(
+    Object.is(boundNegativeZero, -0),
+    "Failed to bind correct value for -0.",
+  );
+
+  const [[boundPositiveMassive]] = db.query("SELECT ?", [+20e20]);
```
dyedgreen, line 863: These should also be done against a proper table (see previous. comment) 😅

dyedgreen (changes requested):
Thank you so much for the PR! Looking pretty good so far. One thing I would like to change is to drop the `-0` case.

SQLite serializes REAL values in tables in such a way that the `-0` becomes a `+0` regardless, so the check does not really do anything in practice. 😅

I left a couple of comments on how to structure the test cases as well. Happy to merge once those are addressed. 😅

src/db.ts
jeremyBanks: Great point, done.

I wasn't thinking about `-0` at first, but noticed while writing the tests it seemed to be preserved when `SELECT`ing it, so I decided to include it. I didn't think to check that it was actually preserved when persisted to a real database. Not much point adding this special case if it's not.

test.ts
jeremyBanks: Sure, done.

test.ts
jeremyBanks: Removed, done.

test.ts
jeremyBanks: Removed, done.

test.ts
jeremyBanks: Sure, done as suggested.

test.ts
jeremyBanks: Done

dyedgreen:
LGTM 😍

Thank you @jeremyBanks  for flagging this issue and the quick fix!

dyedgreen (approved):

PR #95, opened 2020-11-20
Use GitHub Actions to verify that generated files are up-to-date

jeremyBanks:
As suggested by @dyedgreen at https://github.com/dyedgreen/deno-sqlite/pull/88#issuecomment-729801045, here is an initial implementation of rebuilding generated files on GitHub actions, instead of trusting binary WASM files that are locally built and committed.


dyedgreen:
Thanks for looking into this @jeremyBanks 😍

Just had a quick look overt this, and had an idea for what to do regarding finding if the files need to be updated: I believe the build does not take that long (assuming the SQLite amalgamation does not need to be rebuild, which should happen almost never), so I think it could be a good idea to just rebuild every time and check if the resulting binary/ js is changed (and commit any changes)

I'll have a more thorough look through this later, but looking good so far 😃

jeremyBanks:
@dyedgreen That makes sense! I was originally concerned that there would be non-determinism in the build that would add spurious changes on a rebuild, but that doesn't appear to be the case, the output seems to be stable, so the easier approach of always rebuilding should be good.

dyedgreen:
> I was originally concerned that there would be non-determinism in the build

That should not be the case, as long as you use the same compiler version, the build *should* be stable, and if not something is up 😅

jeremyBanks:
Now: it rebuilds on every commit, before running tests. If running in response to a `push` event, the action will push the changes back to the original branch and mark the action as a success. If we're running in response to a `pull_request` event, I think we don't have the original branch context to push back to, so instead we mark the build as a failure since we can't verify it.

This is a little bit weird if you push changes to a pull request that's already in progress:  the `push` event will run in my repository, push the changes, and be successful,
image
but the action will also run in the pull request and fail, and the push from Actions won't trigger another Actions run so the updated commit won't have a status:
image

 I'm not sure there's anything that can really be done about that, except to encourage users to build changes locally/in their branch before sending the PR. Keeping bad binaries from slipping in is probably more important than this inconvenience.

Regarding repeatedly hitting the CDNs: unless it's download a ton of data, I'm not too worried about loading resources from other GitHub URLs. The SQLite package download isn't on GitHub, but it's only ~10MB, so I think we can probably get away with that too.

jeremyBanks:
(If you merge, you should probably squash-merge since I'm cluttering this PR with test commits for itself. :P)

jeremyBanks:
Converting back to draft because I think I have caching of the Deno installation and dependencies working, and will add that shortly.

jeremyBanks:
Nevermind, reverted that. Installing those Deno dependencies takes only two seconds without caching, it's not worth adding the complexity.

.github/workflows/tests.yml
```diff
@@ -1,17 +1,74 @@
...
+        rm -f build/sqlite.{js,wasm} build/lib/sqlite3.{c,h}
+        sudo apt-get -y install --no-install-recommends tclsh gcc
+        cd build
+
+        make setup
+        make amalgamation
+        make release
+
+        git status
+        declare had_changes="$(
```
dyedgreen, line 28: I think it might be worth looking into https://github.com/stefanzweifel/git-auto-commit-action for this? This is a lot of bash script here, and I think using that action we can simplify this to something like what he suggests in his README:

```yaml
on:
  pull_request:
  push:
    branches:
      - "main"

jobs:
  build-and-test:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
      with:
        ref: ${{ github.head_ref }}

    - name: Install Deno
      // etc...

    - uses: stefanzweifel/git-auto-commit-action@v4
      with:
        commit_message: (GitHub Action) Rebuild SQLite WASM
```

dyedgreen: (I believe (but we need to test this), that this should also work on PRs/ that there is some way to make it work on PRs)

.github/workflows/tests.yml
```diff
@@ -1,17 +1,74 @@
...
+  build-and-test:
     runs-on: ubuntu-latest
     steps:
+    - uses: actions/checkout@v2
+
     - name: Install Deno
       run: curl -fsSL https://deno.land/x/install/install.sh | sh
-    - uses: actions/checkout@v1
+
+    - name: Build generated files
```
dyedgreen, line 14:
```suggestion
    - name: Build SQLite
```

dyedgreen (changes requested):
Thank you again for submitting the PR (and for bearing with me for a review for so long; hopefully you had a chance to relax over Christmas 🎄)!

I like the general direction, but I want to try to see if we can make it work on PRs as well (and I think by using an existing action we can simplify the bash script sorcery needed quite a bit ;)

jeremyBanks:
I may not have time to get back to this soon.

dyedgreen:
No worries, there is no rush. Apologies again for keeping you waiting during Christmas 😅😬

build/hack/bundle.js
```diff
@@ -5,19 +5,20 @@ const wasm = await Deno.readFile(src);
...
 }
 
 await Deno.writeFile(
   dest,
   new TextEncoder().encode(
     `import env from "./vfs.js";
 
-const wasm =
-  "${encode(wasm)}";
+const wasm = \`
```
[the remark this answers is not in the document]
jeremyBanks: reverting changes to bundle.js

.github/workflows/tests.yml
```diff
@@ -1,17 +1,74 @@
...
+      id: build
+      run: |
+        export DENO_INSTALL="/home/runner/.deno"
+        export PATH="$DENO_INSTALL/bin:$PATH"
+        rm -f build/sqlite.{js,wasm} build/lib/sqlite3.{c,h}
```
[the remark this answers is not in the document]
jeremyBanks: removing

.github/workflows/tests.yml
jeremyBanks: done

jeremyBanks:
@dyedgreen Done, but I think it won't run on this pull request itself until it's merged into `master`, so I can't test that part.

.github/workflows/generate.yml
```diff
@@ -0,0 +1,30 @@
+on:
+  pull_request:
```
dyedgreen, line 2: Since the action won't run on PRs anyways, why not just have it run on push to master? 😅 
```suggestion
```

.github/workflows/generate.yml
```diff
@@ -0,0 +1,30 @@
+on:
```
dyedgreen, line 1: Can we name this file `wasm.yml`? I feel like `generate` is too generic 😅

dyedgreen (changes requested):
Thank you so much for taking the time to see this through @jeremyBanks!

I think we're pretty much ready to merge this. (I had two very minor points about the workflow file name, and removing the PR trigger, but other than that it should be good to go now 🥳 )

.github/workflows/generate.yml
jeremyBanks: done

.github/workflows/generate.yml
jeremyBanks: done

dyedgreen (approved):
LGTM! 😍 🥳

jeremyBanks:
Gah. It looks like it didn't verify that the workflow was valid because it didn't run until after merging. >_<

I think that should have been

```yaml
file_pattern: build/sqlite.js build/sqlite.wasm
```

instead.

dyedgreen:
No worries. I noticed GH Actions was complaining, so I made a few commits to get it to run. I really hate tinkering with CI, since it's so painful to test :/

dyedgreen:
Yes, it seems to run now 🥳

PR #96, opened 2020-11-21
Add tclsh, gcc, and libc6-dev (required to rebuild SQLite) to dev Dockerfile

jeremyBanks:
As suggested at https://github.com/dyedgreen/deno-sqlite/pull/88#issuecomment-729801045, here is a separate PR updating the dev container Dockerfile to add the dependencies necessary to rebuild SQLite from source.


dyedgreen (approved):
LGTM 😃

PR #99, opened 2021-01-04
Update README.md

Dialga:
Add disadvantage


dyedgreen:
Thanks for opening the PR @Dialga!

This used to be how the module operates. However, newer versions of the module now use a custom VFS, which let's SQLite directly read and write from disc as needed. (See https://github.com/dyedgreen/deno-sqlite/blob/master/build/src/vfs.c)

Dialga:
Oh? Sorry I didn't know that, maybe get the other project to remove their statement then?
Thanks

dyedgreen:
Oh, thanks for flagging that up! I’ll have a look into that. (I’m assuming the other project you mean is the plug-in based version of SQLite?)

PR #101, opened 2021-01-19
Add "workflow_dispatch:" to wasm.yml

jeremyBanks:
This will allow us to manually rebuild it on our branches when we want to.

### DIFF
--- a/.github/workflows/wasm.yml
+++ b/.github/workflows/wasm.yml
@@ -1,6 +1,7 @@
 name: wasm
 
 on:
+  workflow_dispatch:
   push:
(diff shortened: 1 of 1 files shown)


jeremyBanks:
I think this will let me manually rebuild the WASM files, in my fork, while avoiding the committing-locally-generated files issue.

jeremyBanks:
actually I'll just do this in the other PR, and test it in my fork, one minute

dyedgreen:
It might make sense to just add a build step to the test action? (So it will always just test with the build produced by the committed version)

PR #102, opened 2021-01-19
Re-Build SQLite in test runner

dyedgreen:


jeremyBanks:
seems like an appropriate workaround for https://github.com/stefanzweifel/git-auto-commit-action#commits-of-this-action-do-not-trigger-new-workflow-runs 👍

PR #103, opened 2021-01-22
Fix broken formatting check in Deno 1.7.0

jeremyBanks:
Deno 1.7 includes https://github.com/denoland/deno/pull/8887, which extends `deno fmt` to also include Markdown files. This has caused the current formatting check to break on GitHub Actions.

This PR applies `deno fmt`'s formatting changes to all Markdown files in the repository, but also updates the check so it doesn't complain about the otherwise-formatted Markdown files inside the downloaded SQLite source or the `.git` index.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,7 +18,9 @@ jobs:
         cd build
         make setup amalgamation release
     - name: Run the formatter
-      run: /home/runner/.deno/bin/deno fmt --check --unstable --ignore=build/sqlite-src,.git
+      run: |
+        shopt -s globstar failglob
+        /home/runner/.deno/bin/deno fmt --check **/*.{js,ts}
     - name: Run tests
       run: /home/runner/.deno/bin/deno test --allow-read --allow-write test.ts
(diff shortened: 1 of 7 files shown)


dyedgreen:
Hey, thanks for catching this! I think since we have a bunch of markdown files which come from third parties (SQLite sources), and some which are automatically generated (api docs), I think it doesn’t really make sense to format them with a code formatter.

Do you think it makes sense to just format ts/ js files and ignore markdown ones? (I think you could do something like `deno fmt --check --ignore ./**.md`)

jeremyBanks:
@dyedgreen Done, except that I explicitly specified to include `js`/`ts` files instead of exclude `md` files, because I wasn't able to figure out how to get the shell globbing to work properly for the `--exclude=` argument.

dyedgreen:
Yeah, I think explicitly including makes more sense, since that should be robust to changes in the future 😅

dyedgreen (approved):

dyedgreen:
LGTM, thanks for spotting this 😃

PR #109, opened 2021-03-13
SQLite 3.35.0

jeremyBanks:
<blockquote>
<h2><a href="https://sqlite.org/releaselog/3_35_0.html">SQLite Release 3.35.0 On 2021-03-12</a></h2><p><ol class='lessindent'>
<li> Added <a href="https://sqlite.org/../lang_mathfunc.html">built-in SQL math functions()</a>. (Requires the
<a href="https://sqlite.org/../compile.html#enable_math_functions">-DSQLITE_ENABLE_MATH_FUNCTIONS</a> compile-time option.)
<li> Added support for <a href="https://sqlite.org/../lang_altertable.html#altertabdropcol">ALTER TABLE DROP COLUMN</a>.
<li> Generalize <a href="https://sqlite.org/../lang_upsert.html">UPSERT</a>:
<ol type="a">
<li> Allow multiple ON CONFLICT clauses that are evaluated in order,
<li> The final ON CONFLICT clause may omit the conflict target and
yet still use DO UPDATE.
</ol>
<li> Add support for the <a href="https://sqlite.org/../lang_returning.html">RETURNING</a> clause on <a href="https://sqlite.org/../lang_delete.html">DELETE</a>, <a href="https://sqlite.org/../lang_insert.html">INSERT</a>, and
<a href="https://sqlite.org/../lang_update.html">UPDATE</a> statements.
<li> Use less memory when running <a href="https://sqlite.org/../lang_vacuum.html">VACUUM</a> on databases containing very large
TEXT or BLOB values.  It is no longer necessary to hold the entire TEXT
or BLOB in memory all at once.
<li> Add support for the <a href="https://sqlite.org/../lang_with.html#mathint">MATERIALIZED</a> and <a href="https://sqlite.org/../lang_with.html#mathint">NOT MATERIALIZED</a> hints when
specifying <a href="https://sqlite.org/../lang_with.html">common table expressions</a>.  The default behavior was
formerly NOT MATERIALIZED, but is now changed to MATERIALIZED for
CTEs that are used more than once.
<li> The <a href="https://sqlite.org/../c3ref/c_dbconfig_defensive.html#sqlitedbconfigenabletrigger">SQLITE_DBCONFIG_ENABLE_TRIGGER</a> and <a href="https://sqlite.org/../c3ref/c_dbconfig_defensive.html#sqlitedbconfigenableview">SQLITE_DBCONFIG_ENABLE_VIEW</a>
settings are modified so that they only control triggers and views
in the main database schema or in attached database schemas and not in
the TEMP schema. TEMP triggers and views are always allowed.
<li> Query planner/optimizer improvements:
<ol type="a">
<li> Enhancements to the <a href="https://sqlite.org/../optoverview.html#minmax">min/max optimization</a> so that it works better
with the IN operator and the OP_SeekScan optimization of the 
previous release.
<li> Attempt to process EXISTS operators in the WHERE clause as if
they were IN operators, in cases where this is a valid transformation
and seems likely to improve performance.
<li> Allow UNION ALL sub-queries to be <a href="https://sqlite.org/../optoverview.html#flattening">flattened</a> even if the parent query is a join.
<li> Use an index, if appropriate, on IS NOT NULL expressions in the WHERE clause,
even if STAT4 is disabled.
<li> Expressions of the form "x IS NULL" or "x IS NOT NULL" might be converted to simply
FALSE or TRUE, if "x" is a column that has a "NOT NULL" constraint and is not
involved in an outer join.
<li> Avoid checking foreign key constraints on an UPDATE statement if the UPDATE does
not modify any columns associated with the foreign key.
<li> Allow WHERE terms to be <a href="https://sqlite.org/../optoverview.html#pushdown">pushed down</a> into sub-queries
that contain window functions,
as long as the WHERE term is made up of entirely of constants and copies of expressions
found in the PARTITION BY clauses of all window functions in the sub-query.
</ol>
<li> <a href="https://sqlite.org/../cli.html">CLI</a> enhancements:
<ol type="a">
<li> Enhance the ".stats" command to accept new arguments "stmt" and
"vmstep", causing prepare statement statistics and only the
virtual-machine step count to be shown, respectively.
<li> Add the ".filectrl data_version" command.
<li> Enhance the ".once" and ".output" commands so that if the destination argument
begins with "|" (indicating that output is redirected into a pipe) then the
argument does not need to be quoted.
</ol>
<li> Bug fixes:
<ol type="a">
<li> Fix a potential NULL pointer dereference when processing a
syntactically incorrect SELECT statement with a correlated WHERE
clause and a "HAVING 0" clause.  (Also fixed in the 3.34.1 patch release.)
<li> Fix a <a href="https://www.sqlite.org/src/info/ee51301f316c09e9">bug in the IN-operator optimization</a>
of version 3.33.0 that can cause an incorrect answer.
<li> Fix incorrect answers from the <a href="https://sqlite.org/../lang_expr.html#like">LIKE operator</a> if the pattern ends with "%" and there
is an "ESCAPE '_'" clause.
</ol>
</blockquote>

This PR updates the build to use the new SQLite release, and adds a unit test to verify that the `RETURNING` clause is supported.


jeremyBanks:
WASM Size Change: [642498 Bytes -> 665576 Bytes (+4%)](https://github.com/jeremyBanks/deno-sqlite/commit/845c7e799bb943d9b24d624d32f4259b07db623f)

WASM Size Change with `ENABLE_MATH_FUNCTIONS` (not in this branch): [642498 Bytes -> 695163 Bytes (+8%)](https://github.com/jeremyBanks/deno-sqlite/commit/2b60853e2b78024194d200fc837656e2be41329d)

dyedgreen:
Thanks for doing this @jeremyBanks !

What are your thoughts on enabling the math functions? (I would say to hold off, since the won’t have many users and add an extra 4%, but curious to hear your thoughts before merging 😅)

jeremyBanks:
I could go either way, but I might leave it out for now until somebody expresses a need for it.

jeremyBanks:
I really should be comparing the compressed sizes, but I was lazy because it was annoying to do so. Filed #110 to make that easier going forward.

jeremyBanks:
```
                   master         v3.35           v3.35 w/ math
  sqlite.js      ➜ 857303 bytes ➜ 888075 (+ 4%) ➜ 927523 (+ 8%) 
  sqlite.js.br   ➜ 287766 bytes ➜ 297608 (+ 3%) ➜ 316730 (+10%)
  sqlite.js.gz   ➜ 349031 bytes ➜ 361214 (+ 3%) ➜ 381984 (+ 9%)
  sqlite.wasm    ➜ 642498 bytes ➜ 665576 (+ 4%) ➜ 695163 (+ 8%) 
  sqlite.wasm.br ➜ 226870 bytes ➜ 235437 (+ 4%) ➜ 251301 (+11%) 
  sqlite.wasm.gz ➜ 266541 bytes ➜ 276518 (+ 4%) ➜ 294440 (+10%) 
                                 (% changes relative to master)
```

It looks like the new math code doesn't compress particularly well. I might lean against including it at this time.

dyedgreen:
Thanks for doing this @jeremyBanks, looks good to me, let's bump the version 😃

dyedgreen (approved):

Issue #108, opened 2021-03-07
Enable `fts` (full text search) for Deno-sqlite


dyedgreen:
> I've just run into this too. I was trying to load the `fts5` extension.

_(Originally posted by @hobochild in https://github.com/dyedgreen/deno-sqlite/issues/100#issuecomment-792308703 )_

Full text search seems like a common problem and it might make sense to enable full text search by default (since there is currently no way to import extensions at run-time). But I wanted to see how many others are interested in this functionality. (It it turns out to be too niche, it might make sense to ask people to compile their own WASM module with the extension enabled.)

For reference: [SQLite full text search docs](https://www.sqlite.org/fts5.html)

artisonian:
I'd like to see `fts5` support by default. What about `rtree` (considering it [is also available](https://sqlite.org/cli.html) in the `sqlite` CLI)?

dyedgreen:
Hm. I would be curious to test how much each feature effects the WASM binary size vs how much they are used to better understand the tradeoffs involved.

It might also be interesting to offer some more documentation on how to compile your own version if you have specific needs.

artisonian:
This is anecdotal, but I ran `deno info` after enabling `fts5` and `rtree`:

```
https://deno.land/x/sqlite@v2.3.2/build/sqlite.js (810.68KB)
file:///src/vendor/deno-sqlite/build/sqlite.js (1.03MB)
```

dyedgreen:
A thought on loading sqlite extensions: with `Deno.dlopen` it is possible to run code from a shared object. I’m not sure how much memory sharing SQLite extensions require; but if it’s sufficiently little (we can’t really easily expose the WASM memory to a so I believe), it might be possible to run arbitrary SQLite extensions.

ashersaupingomez:
Having `fts` supported would be amazing. Are there any workarounds in the meantime?

lusbuab:
@ashersaupingomez 

> Having `fts` supported would be amazing. Are there any workarounds in the meantime?


You'd need to compile the WASM module yourself with the necessary build options enabled. https://www.sqlite.org/fts5.html should include all necessary information.

mdkq:
Any update on this? I'm having no problem with a custom build with fts5 enabled but would prefer not to have the repo sitting in my project

dyedgreen:
I've had some thoughts around using denos dll loader / a wash loader set-up to support loading custom extensions, but I don't have any specific progress on that yet.

[...]

PR #110, opened 2021-03-15
Log and commit raw and compressed sizes for sqlite.wasm and sqlite.js after building

jeremyBanks:
To make it easier to assess the size differences between builds, such as is being discussed in https://github.com/dyedgreen/deno-sqlite/pull/109#issuecomment-799163057 and https://github.com/dyedgreen/deno-sqlite/issues/108#issuecomment-794313033, this updates the build actions to log the sizes of `build/sqlite.wasm` and `build/sqlite.js`, raw, gzipped, and brotlied.

Output for current `master` build:
https://github.com/jeremyBanks/deno-sqlite/runs/2114753816?check_suite_focus=true#step:4:2664
```
  sqlite.js      ➜ 857303 bytes
  sqlite.js.br   ➜ 287766 bytes
  sqlite.js.gz   ➜ 349031 bytes
  sqlite.wasm    ➜ 642498 bytes
  sqlite.wasm.br ➜ 226870 bytes
  sqlite.wasm.gz ➜ 266541 bytes
```

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -62,6 +62,13 @@ bundle:
 	$(DENO) run --allow-read --allow-write hack/bundle.js $(OUT_WA) $(OUT_BN)
 	$(DENO) fmt $(OUT_BN)
 
+size.txt: sqlite.js sqlite.wasm
+	gzip --best < sqlite.js > sqlite.js.gz
+	gzip --best < sqlite.wasm > sqlite.wasm.gz
+	brotli --best < sqlite.js > sqlite.js.br
+	brotli --best < sqlite.wasm > sqlite.wasm.br
+	ls -l sqlite.* | awk '{printf "%-15s➜%7s bytes\n",$$9,$$5}' | tee size.txt
+
 debug: FLGS += $(DFLG)
 debug: build
 debug: bundle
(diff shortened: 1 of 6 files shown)


build/Makefile
```diff
@@ -62,6 +62,13 @@ bundle:
 	$(DENO) run --allow-read --allow-write hack/bundle.js $(OUT_WA) $(OUT_BN)
 	$(DENO) fmt $(OUT_BN)
 
+size.txt: sqlite.js sqlite.wasm
```
dyedgreen, line 65: (Github won't let me comment on the non-changed line), but I think this would be valuable to add to the `release` target. That way, it will get run automatically after a release build, so the figures are always up to date and if someone submits a PR, it's likely the will have run it locally.

As part of this as well: some people might not have the brotli cli installed, so I think it should fail gracefully in those cases such that development won't be affected (the CI can then jump in and update the files in the automatic run)

(The second point would only make sense it we decide to add it to the release build though; what do you thin @jeremyBanks?)

dyedgreen (commented):
One small question around whether this should be run with every release build, but I like this idea a lot to track changes in the binary size 😃

build/Makefile
jeremyBanks: Good points, I agree. I will make the changes.

dyedgreen (approved):
LGTM 😄

PR #115, opened 2021-03-19
Better types

### LINKED ISSUE #113: Change query API types to better reflect expectations
Currently, the type signature of `query` looks like this:
```typescript
function query(sql: string, values?: object | QueryParam[]): Rows {}
```

I think something like this would be better:
```typescript
function query(sql: string, values?: Record<string, QueryParam> | QueryParam[]): Rows {}
```

dyedgreen:
Closes #113, and adds type generation for the API exposed from C.

### DIFF
--- a/src/row_objects.ts
+++ b/src/row_objects.ts
@@ -32,8 +32,8 @@ export class RowObjects {
    * https://exploringjs.com/es6/ch_iteration.html#sec_closing-iterators
    */
   return(): IteratorResult<Record<string, any>> {
-    let { value, done } = this._rows.return();
-    return { done, value: {} };
+    this._rows.return();
+    return { done: true, value: null };
   }
 
   /**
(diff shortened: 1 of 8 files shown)


jeremyBanks:
Great enhancement! Neat implementation.

src/row_objects.ts
```diff
@@ -31,23 +31,23 @@ export class RowObjects<T extends any = Record<string, any>> {
...
   /**
    * RowObjects.next
    *
    * Implements the iterator protocol.
    */
-  next(): IteratorResult<T> {
+  next(): IteratorResult<Record<string, any>> {
     const { value, done } = this._rows.next();
     if (done) {
       return { value: value, done: true };
```
dyedgreen, line 47: I think this should not work (`this._rows.next()` should return `any[]`, not `Record<string, any>`, meh TypeScript 😢)

PR #116, opened 2021-03-20
Add test database generator for CI and local testing

### LINKED ISSUE #107: Clean up tests 🧼
- [ ] Add c generation code for big test database (generate it in C by calling a process, then use it in the test)
- [x] Use the permissions query api instead of home cooked solution

dyedgreen:
closes #107 

- [x] Generate large (>2GB) database
- [x] Set up tests to check for the generated database and use it in the large DB test case (check the case fails when the error is present!)
- [x] Set up CI to generate the test DB before running the tests


dyedgreen:
Some observations: This makes the tests much slower to run in CI, it might make sense to look into caching later. Locally, this is perfectly fine; the test db can be built once and reused.

PR #117, opened 2021-03-22
Prepared statements

### LINKED ISSUE #111: Offer API for prepared statements (enables things like more efficient bulk insert)
Would it be possible to have a bulk insert feature?

I looked at the code and it looks like with the `query` method if we pass in an array of objects or array of arrays then we could reuse the statement and get a high performance bulk insert.

dyedgreen:
Closes #111

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -1,6 +1,6 @@
 // @deno-types="../build/sqlite.d.ts"
 import instantiate, { StatementPtr, Wasm } from "../build/sqlite.js";
-import { getSQLiteError, setArr, setStr } from "./wasm.ts";
+import { setArr, setStr } from "./wasm.ts";
 import { Status, Values } from "./constants.ts";
 import SqliteError from "./error.ts";
 import { Empty, Rows } from "./rows.ts";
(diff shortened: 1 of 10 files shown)


src/wasm.ts
```diff
@@ -74,3 +74,11 @@ export function getStr(wasm: Wasm, ptr: number): string {
     return str;
   }
 }
+
+export function getSQLiteError(wasm: Wasm, code?: number): SqliteError {
```
dyedgreen, line 78: Note to self: I feel like this should be the `SqliteError` constructor. Maybe a signature like
```typescript
class SqliteError extends Error {
  constructor(context: Wasm | {code: number, message: string});
}

File: src/db.ts [hunk 2]
```diff
@@ -112,21 +117,103 @@ export class DB {
...
+      default:
+        this._wasm.finalize(stmt);
+        throw new SqliteError(this._wasm, status);
+    }
+  }
+
+  /**
+   * DB.prepareQuery
+   *
+   * TODO(dyedgreen): Write docs.
```
dyedgreen, line 148: Note to self: Don't forget to add these before merging!

Issue #14, opened 2020-01-25
Make benchmarks more visible


dyedgreen:
It would be nice to record bencharks on commit to master to a csv file and render them for display in the READMEs. (compare https://deno.land/benchmarks)

andykais:
very curious how this library's performance compares to the infamous [better-sqlite3](https://github.com/JoshuaWise/better-sqlite3) library

dyedgreen:
Yeah, that might be an interesting comparison. I'll have a look at seeing how they compare (although it should probably be mentioned, that this library stores the files in-memory only, with explicit save, due to limitations in WASM; resulting in an API that has to specifically save changes, similar to Python)

andykais:
> this library stores the files in-memory only

Is this still true after https://github.com/dyedgreen/deno-sqlite/issues/49 was closed?

dyedgreen:
No, the library now provides a custom VFS for Sqlite, which means that files are read and written by Sqlite directly. How much of a file is stored in memory is entirely up to Sqlite, just like you would get when using Sqlite e.g. as a C library 😅

andykais:
Sweet! Fwiw, this is a benchmark comparing inserting 1,000,000 rows inside a transaction with your library vs better-sqlite3. The difference is pretty much negligible

#### deno sqlite:
```ts
import { DB } from "https://deno.land/x/sqlite/mod.ts";


const db = new DB("test.db");

db.query('DROP TABLE IF EXISTS people')
db.query(`CREATE TABLE IF NOT EXISTS people (
  name TEXT NOT NULL
)`)

const rows = Array(1000000).fill(null).map((_,i) => i)
const stmt = db.prepareQuery("INSERT INTO people (name) VALUES (?)")
console.time('insert 1,000,000 rows in a transaction')
db.query('begin')
for (const row of rows) {
  stmt([row])
}
console.timeEnd('insert 1,000,000 rows in a transaction')
console.time('select from 1,000,000 rows')
db.query('SELECT COUNT(*) FROM people').next()
db.query('SELECT * FROM people LIMIT 1').next()
console.timeEnd('select from 1,000,000 rows')
db.query('commit')
``` 
insert 1,000,000 rows in a transaction: 1435ms
select from 1,000,000 rows: 17ms



#### better-sqlite:
```js
const Sqlite = require('better-sqlite3')

const db = new Sqlite('test.db')

db.exec(`DROP TABLE IF EXISTS people;
CREATE TABLE people (
[...]

PR #138, opened 2021-07-24
New API for `query` and `prepareQuery`

### LINKED ISSUE #132: Provide specialized query methods for improved performance
The common patterns in other libraries are:
- execute (query none; although inserts are already query none, so just using the existing `query` is probably fine/ has no performance difference)
- query one
- query all
- query cursor / iterator (the current `query`)

This would be a performance only feature; since the existing `query` already fills all the existing cases:
- execute: `db.query(...) // ignores return value`
- query one: `let [firstRow] = db.query(...)`
- query all: `[...db.query(...)]`
- query cursor / iterator: that just the existing `db.query(...)`

_Originally posted by @dyedgreen in https://github.com/dyedgreen/deno-sqlite/issues/14#issuecomment-847397847_

dyedgreen:
Closes #132 

This is a re-design of the API, to make it easier to used, and more performant. Feedback is very welcome!

See the full API [here](https://doc.deno.land/https/raw.githubusercontent.com%2Fdyedgreen%2Fdeno-sqlite%2Frework_api%2Fmod.ts).

cc @jeremyBanks @uki00a @halvardssm @andykais

### To Do
- [x] Finalize API design (write examples to see how the new API works in practice)
- [x] Write docs for new functions
- [x] Update benchmarks to measure performance in a more meaningful way
- [x] Add tests for the new API surface (`query`, `queryAll`, `queryOne`, and `execute` should be tested explicitly)


dyedgreen:
# Performance tests (preliminary)

Trying to run the comparison from: https://github.com/dyedgreen/deno-sqlite/issues/14#issuecomment-847119039

```typescript
// Deno
import { DB } from "./mod.ts";

const selectCount = 13000;
const db = new DB("test.db");

db.query(
  "CREATE TABLE people (id INTEGER PRIMARY KEY AUTOINCREMENT, name TEXT NOT NULL)",
);

db.query("begin");
const insert = db.prepareQuery("INSERT INTO people (name) VALUES (?)");
for (let i = 0; i < selectCount; i++) {
  insert.execute([i]);
}
db.query("commit");

const select = db.prepareQuery("SELECT * FROM people WHERE name = ?");
console.time(`select ${selectCount} times`);
for (let i = 0; i < selectCount; i++) {
  const _row = select.queryOne([i]);
}
console.timeEnd(`select ${selectCount} times`);
```

```javascript
Node / better-sqlite
const db = require("better-sqlite3")("test.db");

const select_stmt = db.prepare("SELECT * FROM people WHERE name = ?");
const select_count = 13000;
console.time(`select ${select_count} times`);
for (let i = 0; i < select_count; i++) {
  select_stmt.get(i);
}
console.timeEnd(`select ${select_count} times`);
```

```
results deno: select 13000 times: 14446ms
results node: select 13000 times: 23.505s
```

src/query.ts
```diff
@@ -0,0 +1,526 @@
...
+import { StatementPtr, Wasm } from "../build/sqlite.js";
+import { getStr, setArr, setStr } from "./wasm.ts";
+import { Status, Types, Values } from "./constants.ts";
+import { SqliteError } from "./error.ts";
+
+/**
+ * The default type for returned rows.
+ */
+// deno-lint-ignore no-explicit-any
+export type Row = Array<any>;
```
dyedgreen, line 11: Should this be `type Row = Array<unknown>` instead?

dyedgreen: (The benefit being that it encourages people to type-annotate their queries)

dyedgreen:
A thought about the api: most people will name their prepared query objects 'query'. E.g.
```typescript
const userQuery = db.prepareQuery("…");
```

This can make the names of the query methods awkward to use eg `userQuery.queryOne()`. To make this less awkward they could be renamed to not include 'query'. I would propose:

- `query` -> `iter`
- `queryAll` -> `all`
- `queryOne` -> `one`
- `execute` -> is already good

Which would make the call above `userQuery.one()`.

src/query.ts
andykais: I think its best to stick with `any`. In most cases, we _know_ what the type is, we just have to cast it. Unknown is typically when you _dont_ know the type and need to parse it. 
```ts
const row: PhoneNumber = stmt.queryOne() as any
```

This feels very similar to the discussion that the official typescript repo had regarding `JSON.parse`. Ultimately they chose to keep it returning `any`, and let developers wrap it in another function if they wished to use `unknown`

andykais:
I'm not so sure those benchmarks are accurate. I ran the two benchmarks (modifying them slightly so nodejs doesnt depend on deno sqlite to instantiate the db) and better-sqlite3 performs faster:

#### better-sqlite3
<details>
<summary>benchmark code</summary>

```ts
const DB = require('better-sqlite3')

const row_count = 13000;
const db = new DB("test.db");

db.exec("DROP TABLE people");
db.exec("CREATE TABLE people (id INTEGER PRIMARY KEY AUTOINCREMENT, name TEXT NOT NULL)");

console.time(`insert ${row_count} times`);
db.transaction(() => {
  const insert = db.prepare("INSERT INTO people (name) VALUES (?)");
  for (let i = 0; i < row_count; i++) {
    insert.run([i]);
  }
})()
console.timeEnd(`insert ${row_count} times`);

const { count } = db.prepare('SELECT COUNT(1) as count FROM people').get()
console.log(`${count} rows in table`)

const select = db.prepare("SELECT * FROM people WHERE name = ?");
console.time(`select ${row_count} times`);
for (let i = 0; i < row_count; i++) {
  const _row = select.get([i]);
}
console.timeEnd(`select ${row_count} times`);
```
</details>

```
insert 13000 times: 25.069ms
13000 rows in table
select 13000 times: 17.103s
```
#### deno sqlite
<details>
<summary>benchmark code</summary>

```ts
import { DB } from "https://raw.githubusercontent.com/dyedgreen/deno-sqlite/rework_api/mod.ts";

const row_count = 13000;
const db = new DB("test.db");

db.query("DROP TABLE people");
db.query("CREATE TABLE people (id INTEGER PRIMARY KEY AUTOINCREMENT, name TEXT NOT NULL)");

console.time(`insert ${row_count} times`);
db.query("begin");
const insert = db.prepareQuery("INSERT INTO people (name) VALUES (?)");
for (let i = 0; i < row_count; i++) {
  insert.execute([i]);
}
db.query("commit");
console.timeEnd(`insert ${row_count} times`);

const count = db.query('SELECT COUNT(1) FROM people')
console.log(`${count} rows in table`)

const select = db.prepareQuery("SELECT * FROM people WHERE name = ?");
console.time(`select ${row_count} times`);
for (let i = 0; i < row_count; i++) {
  const _row = select.one([i]);
}
console.timeEnd(`select ${row_count} times`);
```
</details>

```
insert 13000 times: 76ms
13000 rows in table
select 13000 times: 27486ms
```

I'd like to point out that this pr is still a huge win for this library, ergonomically and speedwise, better-sqlite3 just may have a few more performance tricks up its sleeve is all 🙂

dyedgreen:
Right, if you so the insert and select in the same connection sqlite will cache some of it, so better-sqlite is still faster. But the speed up for us is still significant 😅

src/query.ts
dyedgreen: Hm the thing with the new api is that you can easily annotate the row the type, which is preferable especially for prepared queries since it’s that means you can annotate the type next to the query and then have it checked everywhere the query is used:

```typescript
const query = dB.prepareQuery<[number, string]>("SELECT id, name FROM users");
// …
for (const [id, name] of query.iter()) {
// …
}
```

My hope was that using unknown would encourage annotating the type, which would then mean the type annotation is always close to where the query SQL is defined as well 🤔

dyedgreen: (or type annotation for the top level query method without a prepare)
```typescript
const users = db.query<[number, string]>("SELECT id, name FROM users");
```

dyedgreen: @andykais I was trying to find the TypeScript discussion you mentioned, but instead I stumbled across [this thread](https://github.com/microsoft/TypeScript/issues/10715) around `unknown`, which cites `JSON.parser` as one of the motivations for the `unknown` type.

Especially since the API we provide has an explicit type parameter for annotating `query` and `preparedQuery`, I think using `unknown` is a win overall, since it's the safer default and encourages co-locating type definitions with the query they depend on (vs having `any` put the burden on annotating/ checking the type everywhere the prepared query is used).

andykais:
one neat typing suggestion would be to tie a table to a prepared statement. Then statements could be reused without the need to cast variables each time. E.g.
```ts
interface PhoneNumber {
  id: string
  telephone: number
}

// type param would default to any of course (db.prepareQuery<any>)
const phone_number_select = db.prepareQuery<PhoneNumber>('SELECT * FROM phone_number')

// row is of type 'PhoneNumber' without manually casting it
const row phone_number_select.one()
```

essentially all this does is push the type casting up from the query execution to the statement prepares.
```ts
// without typed statements, we need to cast on each query execute
const row: PhoneNumber phone_number_select.one()
```

dyedgreen:
> one neat typing suggestion would be to tie a table to a prepared statement. Then statements could be reused without the need to cast variables each time.

That’s already how it works: Both `query` and `prepareQuery` accept a type parameter which you can set to reflect your select statements:
```typescript
const emails = db.query<[string]>("SELECT email FROM users");

const query = db.prepareQuery<[string, number]>("SELECT name, age FROM users WHERE country = :country");
for (const [name, age] of query.iter({ country }) {
  // …
}
```

PR #142, opened 2021-07-30
Provide an API for returning rows as objects containing key value pairs

### LINKED ISSUE #140: return rows as [column: value] objects rather than value arrays
Unsure if this should be a part of #138 (apologies for suggesting so many changes 😅) but it would probably be very useful to return rows as key value pairs of column names and values, rather than what they are now, which is an array of values. E.g.
```ts
const row = person_select.one()
console.log(row.name)
console.log(row.age)
```

dyedgreen:
Closes #140

### DIFF
--- a/src/query.ts
+++ b/src/query.ts
@@ -349,6 +349,10 @@ export class PreparedQuery<
     return this as RowsIterator<R>;
   }
 
+  /**
+   * Like `iter` except each row is returned
+   * as an object containing key-value pairs.
+   */
   kvIter(params?: QueryParameterSet): RowsIterator<O> {
     this.iter(params);
     this._iterKv = true;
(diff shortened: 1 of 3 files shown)


dyedgreen:
I am still not 100% sure on the naming scheme for the methods. Currently, this uses `kvXxx`, but there are a few possible alternatives I considered, using `all` as an example:
```
allObj
allKv
allKeyValue
allObjects
allRec
allRecords

objAll
kvAll
recAll
recordsAll
objectsAll
keyValueAll

oall
kvall
```

src/query.ts
```diff
@@ -278,6 +290,24 @@ export class PreparedQuery<R extends Row = Row> {
     return (row as unknown) as R;
   }
 
+  private makeRowObject(row: Row): O {
+    if (this._rowKeys == null) {
+      const rowCount = this._wasm.column_count(this._stmt);
```
andykais, line 295: sorry, I know this pr is already closed but are we always making two database calls when we grab row objects? Is there not a way to return all this information in a single call?

dyedgreen: You have to do this at least once to collect the rows names, but the names are memoized for subsequent calls.

PR #156, opened 2021-10-31
Export SqliteError as value

jsejcksn:
Currently, a compiler error is produced when trying to use `SqliteError` as a value:

> 'SqliteError' cannot be used as a value because it was exported using 'export type'.deno-ts(1362)

It is useful to provide the export as a value for cases like the following:

```ts
import {DB, SqliteError} from 'https://deno.land/x/sqlite/mod.ts';

/*...*/

try {
  db.query(/*...*/); // or any other import that might throw an instance of `SqliteError`
}
catch (ex) {
  if (ex instanceof SqliteError) {
    // handle SQLite error
  }
  else { /*...*/ }
}
```

This PR exports the class as a value to address this. (It can, of course, still be used as a type.)

### DIFF
--- a/mod.ts
+++ b/mod.ts
@@ -1,4 +1,5 @@
 export { DB } from "./src/db.ts";
+export { SqliteError } from "./src/error.ts";
 export { Status } from "./src/constants.ts";
 
 export type { SqliteOptions } from "./src/db.ts";
@@ -10,4 +11,3 @@ export type {
   Row,
   RowObject,
 } from "./src/query.ts";
-export type { SqliteError } from "./src/error.ts";


dyedgreen:
I was under the impression that that was already possible with typescript, but it makes sense that it wouldn’t be 😅

Looks good to me though 😊

dyedgreen:
I was under the impression that that was already possible with typescript, but it makes sense that it wouldn’t be 😅

Looks good to me though 😊

jsejcksn:
Thanks. The reason for that is [type-only imports and exports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export) are elided by default (and therefore not available) in the compiled JavaScript.

Will this be released soon (and become available on deno.land/x), or should I just use [the latest wasm build commit branch](https://github.com/dyedgreen/deno-sqlite/tree/f3662458ed7415681144b04a828ff6bb0194ee45) in my imports?

```ts
import {SqliteError} from 'https://raw.githubusercontent.com/dyedgreen/deno-sqlite/f3662458ed7415681144b04a828ff6bb0194ee45/mod.ts';
```

dyedgreen:
I'm happy to make a point release 😅

jsejcksn:
I see [it](https://github.com/dyedgreen/deno-sqlite/releases/tag/v3.1.2) — thanks!

PR #164, opened 2021-12-11
Support transactions

### LINKED ISSUE #162: Add API support for transactions
I think adding first class support for transactions to the API would be nice.

I’m thinking probably a closure based API would be best?

dyedgreen:
Closes #162

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -196,6 +196,15 @@ export class DB {
     return new PreparedQuery<R, O, P>(this._wasm, stmt, this._statements);
   }
 
+  /**
+   * Run a function within the context of a database
+   * transaction. If the function throws an error,
+   * the transaction is rolled back. Otherwise, the
+   * transaction is committed when the function returns.
+   *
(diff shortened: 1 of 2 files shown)


dyedgreen:
One open question is whether this should support async functions inside the transaction. I'm tempted to use an API similar to the standard libraries:
```typescript
async function transaction(closure: () => Promise<V>): Promise<V>
function transactionSync(closure: () => V): V
```

But there is a big open question there what should happen when you try to run transactions in parallel, e.g.:
```typescript
let first = db.transaction(async () => {
  / * run queries ... * /
});
let second = db.transaction(async () => {
  / * run queries ... * /
});
let [firstResult, secondResult] = Promise.all(first, second);
```

Should these transactions be serialized, or should they be nested (and if they are nested, do we guarantee a nesting order)?

Because of this difficulty with the interface, I'd be tempted to say transactions must be synchronous, and if you need to do some async work, conditional on a query result you must run in a transaction, there is always `BEGIN TRANSACTION`, which gives you more fine grained control.

PR #168, opened 2022-01-02
Experimental browser support

dyedgreen:
Support for using `deno-sqlite` in web-browsers, backed by indexedDB


dyedgreen:
This currently doesn’t work in Chrome. I think to properly support all browsers, we need to do the compilation + instantiation async.

This will need some special-casing (init compile, to use with instantiate VS init compile + instantiate + grab instance if ready).

README.md
```diff
@@ -76,6 +76,24 @@ this module seems to no longer be actively maintained.
   SQLite can't acquire file locks or memory map files (e.g. this module does not
   support WAL mode)
 
+## Browser Version (Experimental)
+
+There is **experimental** support for web browser support. You can generate a
```
dyedgreen, line 81: Ugh 😑

browser/mod.ts
```diff
@@ -0,0 +1,42 @@
...
+import { compile } from "../build/sqlite.js";
+
+export { SqliteError } from "../src/error.ts";
+export { Status } from "../src/constants.ts";
+
+/**
+ * It is a bug to use the other functions before the
+ * promise returned by this resolves.
+ */
+export async function init(): Promise<void> {
```
dyedgreen, line 12: This is idenpotent; no need to expose it!

browser/vfs.js
```diff
@@ -0,0 +1,206 @@
...
+    const request = store.get(fileName);
+    request.onsuccess = () => resolve(request.result);
+    request.onerror = () => reject(request.error);
+  });
+  if (file != null && !LOADED_FILES.has(fileName)) {
+    const buffer = new Buffer(file.data);
+    LOADED_FILES.set(fileName, buffer);
+    return buffer;
+  } else {
+    return null;
```
dyedgreen, line 98: Return buffer if already loaded as well!

build/hack/bundle.js
```diff
@@ -29,15 +30,32 @@ function decode(base64) {
...
+export async function compile() {
+  moduleOrInstance.module = await WebAssembly.compile(decode(wasm));
+}
+
+export async function instantiateBrowser() {
   const placeholder = { exports: null };
-  const instance = new WebAssembly.Instance(module, env(placeholder));
+  const instance = await WebAssembly.instantiate(moduleOrInstance.module, env(placeholder));
   placeholder.exports = instance.exports;
-  instance.exports.seed_rng(Date.now());
```
dyedgreen, line 39: We should seed the RNG here as well.

PR #176, opened 2022-02-18
feat: add build support for Windows via MinGW

kyuzu:


build/Makefile
```diff
@@ -6,16 +6,22 @@ OUT_WA = "sqlite.wasm"
 OUT_BN = "sqlite.js"
 OUT_TY = "sqlite.d.ts"
 
+HOST_OS = linux
```
dyedgreen, line 9: Can we move this down to be with the if statement 😅

dyedgreen:
This looks really neat! One thing before we merge this: since I don’t have a windows machine myself to test this, could we [run CI tests on windows](https://docs.github.com/en/actions/using-github-hosted-runners/customizing-github-hosted-runners) as well?

build/Makefile
kyuzu: You mean like the following?
```mk
ifeq ($(OS),Windows_NT)
    # ...
endif

ifeq ($(shell uname),Linux)
    # ...
endif

ifeq ($(shell uname),Darwin)
    # ...
endif
```

kyuzu:
@dyedgreen Sure thing. But, we need WASI SDK [v14](https://github.com/WebAssembly/wasi-sdk/releases/tag/wasi-sdk-14) in order for this to work. The current version, v11, produces some error related to `libstc++-6` during compilation. The problem is with `clang` itself and not the code.

dyedgreen:
I’m happy to upgrade to the latest compiler if it doesn’t break anything on linux/mac 😅

kyuzu:
Alright, I'll work on the CI tomorrow since I'm quite busy today.

dyedgreen:
I think we probably want to use a [build matrix](https://docs.github.com/en/actions/using-jobs/using-a-build-matrix-for-your-jobs). Then we can just maintain one set of steps to run on Linux/Mac/Windows 😅

kyuzu:
Should be fixed now.\

dyedgreen:
I’ll have a looo over this tomorrow, but looks good so far! 😅

kyuzu:
Any updates?
@dyedgreen

.github/workflows/tests.yml
```diff
@@ -4,36 +4,69 @@ on: [push, pull_request]
...
+      run: |
+        export DENO_INSTALL="$USERPROFILE/.deno"
+        export PATH="$DENO_INSTALL/bin:$PATH"
+        cd build
+        make setup amalgamation release
+
     - name: Set Timezone to Non-UTC
-      run: sudo timedatectl set-timezone America/New_York
-    - name: Generate Test Database
+      uses: zcong1993/setup-timezone@master
```
dyedgreen, line 51: This doesn't work for non linux runners (also, I don't think we should pull in a dependency, just do do an if-statement 😅)

We can do something like:
```yaml
- name: Set Non UTC Timezone (Linux)
  if: runner.os == 'Linux'
  run: sudo timedatectl set-timezone America/New_York
- name: Set Non UTC Timezone (Mac)
  if: runner.os == 'macOS'
  run: sudo systemsetup -settimezone America/New_York
- name: Set Non UTC Timezone (Windows)
  if: runner.os == 'Windows'
  run: tzutil /s "Central America Standard Time"
```
(Not 100% sure about the windows one, but let's run it and see)

.github/workflows/tests.yml
```diff
@@ -4,36 +4,69 @@ on: [push, pull_request]
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [windows-latest, ubuntu-latest, macOS-latest]
+        deno-version: [1.19, 1.17, 1.15]
```
dyedgreen, line 11: Can we do instead
```suggestion
        deno-version: [v1.x, canary]
```

.github/workflows/tests.yml
```diff
@@ -4,36 +4,69 @@ on: [push, pull_request]
...
+        make setup amalgamation release
+
     - name: Set Timezone to Non-UTC
-      run: sudo timedatectl set-timezone America/New_York
-    - name: Generate Test Database
+      uses: zcong1993/setup-timezone@master
+      with:
+        timezon: America/New_York
+
+    - name: Generate test DB
```
dyedgreen, line 55:
```suggestion
    - name: Generate Test Database
```

.github/workflows/tests.yml
```diff
@@ -4,36 +4,69 @@ on: [push, pull_request]
...
+
+    - name: Setup MinGW environment
+      if: ${{ matrix.os == 'windows-latest' }}
+      uses: msys2/setup-msys2@v2
+      with:
+        msystem: MINGW64
+        update: true
+        install: git gcc make mingw-w64-x86_64-cc
+
+    - name: Build SQLite (Windows)
```
dyedgreen, line 41: We don't need this step to be separate from the linux build, since `setup-deno` sets up the path for us.

So can delete the `export DENO_INSTALL`/`export PATH` statements from the command, delete the windows specific version, and move the `Setup MinGW Environment` step before the build step.

.github/workflows/tests.yml
```diff
@@ -4,36 +4,69 @@ on: [push, pull_request]
...
       id: build
+      if: ${{ matrix.os != 'windows-latest' }}
       run: |
         export DENO_INSTALL="/home/runner/.deno"
         export PATH="$DENO_INSTALL/bin:$PATH"
-        sudo apt-get -y install --no-install-recommends tclsh gcc brotli
         cd build
         make setup amalgamation release
+
+    - name: Setup MinGW environment
```
dyedgreen, line 33:
```suggestion
    - name: Setup MinGW Environment
```

build/Makefile
```diff
@@ -9,13 +9,27 @@ OUT_TY = "sqlite.d.ts"
...
 
+ifeq ($(shell uname),Darwin)
+	HOST_OS = macos
+endif
+
+ifeq ($(shell uname),Linux)
+	HOST_OS = linux
+endif
+
+# Check if HOST_OS is not set
```
dyedgreen, line 24: This is pretty obvious without a comment 😅 
```suggestion
```

build/Makefile
```diff
@@ -100,7 +114,6 @@ amalgamation:
 
 dlsqlite:
 	curl "$(SQLITE_DLD)" -o "sqlite_dl.zip"
-	sed -n '/sqlite_dl.zip/p' ".checksums" | shasum -c -
```
dyedgreen, line 103: Don't delete the checksum check please

build/Makefile
```diff
@@ -109,10 +122,10 @@ dlsqlite:
 
 dlwasi:
 	curl -L "$(WASI_DLD)" -o "$(WASI_TAR)"
-	sed -n '/${WASI_TAR}/p' ".checksums" | shasum -c -
+	# sed -n '/${WASI_TAR}/p' ".checksums" | shasum -c -
```
dyedgreen, line 125: Same as above. Can you update the checksums for the different versions of the sdks, instead of deleting the check.

dyedgreen (changes requested):
Sorry for the delay, had a pretty busy week.

Except for a few minor comments, this is looking really good! 😊

PR #178, opened 2022-02-21
Update alternatives to point to x/sqlite3

waynebloss:
Hello, the current readme points to an old unmaintained native Deno plugin ([crabmusket/deno_sqlite_plugin](https://github.com/crabmusket/deno_sqlite_plugin)) but now we have [denodrivers/sqlite3](https://github.com/denodrivers/sqlite3) (published at [x/sqlite3](https://deno.land/x/sqlite3)) and it looks like this one is being maintained for now...

One caveat - to use it, you have to run with the `--unstable` flag.

So, I was just wondering if we could update the README to point to that?


README.md
```diff
@@ -56,8 +56,8 @@ db.close();
 If you want something that just works (and is fast), use this library.
 
 Depending on your specific needs, there is also
-[deno_sqlite_plugin](https://github.com/crabmusket/deno_sqlite_plugin), however
-this module seems to no longer be actively maintained.
+[deno_sqlite](https://github.com/denodrivers/sqlite3), however
+this module requires the `--unstable` flag currently.
```
dyedgreen, line 60:
```suggestion
using this module requires the `--allow-ffi` and `--unstable` flag.
```

dyedgreen:
Thank you for pointing this out! Ive left one small wording suggestion, let me know what you think 😊

waynebloss:
Oh, I didn't realize that I needed the `--allow-ffi` flag too probably because I just ran it with `-A` when I tested it. Thanks!

waynebloss:
The necessary flags aren't even noted in their documentation though, so if I use it I'll probably be asking them to add this info too!

PR #182, opened 2022-03-11
Add helper for running multiple statements

wirelyre:
Hello!

I wanted something like Python's [`executescript`](https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.executescript) so I looked into porting it.

It turns out that SQLite already has a convenience method for this, [`sqlite3_exec`](https://sqlite.org/c3ref/exec.html).

If this had been in the interface, I would have saved quite a bit of time debugging. So I hope this will be merged eventually!

## Things left to do
* Rebuild the WASM
* Document the method better
* Add test?
* Wrap in `TRANSACTION`?

Python's implementation looks like [this](https://github.com/python/cpython/blob/5b1b9eacb92dd47d10793a8868246df6ea477ed6/Modules/_sqlite/cursor.c#L713-L777). I think it's complicated because of threading. It also runs `COMMIT` first, but I don't think this library has an auto-commit feature.

I didn't commit the WASM myself because my computer didn't make an identical build on `master` and I didn't want to do it wrong.

### DIFF
--- a/build/sqlite.d.ts
+++ b/build/sqlite.d.ts
@@ -19,7 +19,7 @@ export interface Wasm {
   finalize: (stmt: StatementPtr) => number;
   reset: (stmt: StatementPtr) => number;
   clear_bindings: (stmt: StatementPtr) => number;
-  run_multiple: (sql: StringPtr) => number;
+  exec: (sql: StringPtr) => number;
   bind_int: (stmt: StatementPtr, idx: number, value: number) => number;
   bind_double: (stmt: StatementPtr, idx: number, value: number) => number;
   bind_text: (stmt: StatementPtr, idx: number, value: StringPtr) => number;
--- a/build/sqlite.d.ts
+++ b/build/sqlite.d.ts
@@ -19,6 +19,7 @@ export interface Wasm {
   finalize: (stmt: StatementPtr) => number;
   reset: (stmt: StatementPtr) => number;
   clear_bindings: (stmt: StatementPtr) => number;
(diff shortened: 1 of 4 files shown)


dyedgreen (changes requested):
This is awesome 🤩

I have left a few small nits/bikeshed comments about the naming, but I think generally, this looks really good!

Regarding your things left to do:
- You don't have to commit it / shouldn't commit it, since we have CI to auto-rebuild and commit an updated version, so whatever you do get's overwritten anyways 😄
- Re documenting the method better: I've left a few code review comments
- We should add a test, but it doesn't have to be too fancy. Can we add one which does something like:
   * Create a table; insert some rows in one `execute`, then load back the rows and assert on them
   * Have an `execute` which fails to parse/has some error and assert it throws
- I don't think we should wrap it in an implicit transaction. If you don't want a transaction, this makes the API unusable, but if you want one it's easy enough to just add `BEGIN TRANSACTION` to the SQL / call the method from within `db.transaction()`

wirelyre:
Great review! :-)

wirelyre:
Sorry, I tried to rename the branch but GitHub didn't like it.

dyedgreen (commented):
LGTM 💯

dyedgreen (approved):

dyedgreen:
Awesome work, thank you so much! 😊

I'll merge it once ci passes

wirelyre:
Thanks!

PR #184, opened 2022-03-19
Fix doc for queryEntries

patrickpang:
Fix example code in doc for queryEntries


dyedgreen (approved):
Thanks for spotting this!

PR #203, opened 2022-08-25
Upgrade SQLite and WASI (compiler)

dyedgreen:
Bump the SQLite version to `3.39.2` and the WASI compiler version to `16.0`


Melkaz:
Sorry to bother with such a dumb question: is there a way to show the version of SQLite bundled with this package ?

Thanks,

PR #221, opened 2022-11-04
Add API to determine auto commit mode

rherrmann:
Add an `autoCommit` property to `DB` that forwards the result of `sqlite3_get_autocommit`.

### DIFF
--- a/build/src/wrapper.c
+++ b/build/src/wrapper.c
@@ -112,6 +112,11 @@ double EXPORT(total_changes) () {
   return (double)sqlite3_total_changes(database);
 }
 
+// Returns whether in auto commit mode
+int EXPORT(autocommit) () {
+  return sqlite3_get_autocommit(database);
+}
+
 // Wraps sqlite3_prepare. Returns statement id.
 sqlite3_stmt* EXPORT(prepare) (const char* sql) {
   // Prepare sqlite statement
--- a/src/db.test.ts
+++ b/src/db.test.ts
@@ -163,6 +163,17 @@ Deno.test("database has correct changes and totalChanges", function () {
   assertEquals(6, db.totalChanges);
 });
 
+Deno.test("autoCommit property reflects database state", function () {
+  const db = new DB();
+  assertEquals(db.autoCommit, true);
+
+  db.query("BEGIN");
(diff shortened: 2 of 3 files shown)


src/db.test.ts
```diff
@@ -533,3 +533,20 @@ Deno.test(
     db.close();
   },
 );
+
+Deno.test("autoCommit", {
```
dyedgreen, line 537: We should use a more descriptive name like “autoCommit property reflects database state”. Also, can we move the test to be collocated next to the tests for `changes` and `totalChanges` 😅

src/db.test.ts
```diff
@@ -533,3 +533,20 @@ Deno.test(
...
+  ignore: !TEST_DB_PERMISSIONS,
```
dyedgreen, line 538: We can drop these, since the test doesn’t need to use a database file on disk.

src/db.test.ts
```diff
@@ -533,3 +533,20 @@ Deno.test(
...
+  const db = new DB();
+  assertEquals(db.autoCommit, true);
+
+  db.query("BEGIN");
+  assertEquals(db.autoCommit, false);
+
+  db.query("ROLLBACK");
+  assertEquals(db.autoCommit, true);
+
+  db.close();
```
dyedgreen, line 551: This is not needed for an in memory database 😊

build/src/wrapper.c
```diff
@@ -424,3 +424,7 @@ int EXPORT(deserialize) (const char* schema, void* data, int bytes, int flags) {
   );
   return last_status;
 }
+
+int EXPORT(get_autocommit) () {
```
dyedgreen, line 428: I think calling this just `autocommit` is fine 😅

Also, can we move this next to the functions for changes and total changes + add a short comment about what it does.

dyedgreen (changes requested):
Looks good, thank you so much for contributing 🙌

Some minor stylistic points (and test permissions), and then this should be good to go!

dyedgreen (approved):

dyedgreen:
Awesome! Thank you so much for contributing 🙌

PR #224, opened 2022-12-07
Clarify which module may bypass permissions

cwohlman:
I was initially confused about which module required the `ffi` and `unstable` flags due to the ambiguous use of `this`. Using `that` clearly indicates the other module needs those permissions.


dyedgreen:
That’s a good catch! Thanks for submitting the fix 😊

PR #231, opened 2023-04-22
Release the savepoint on rollback

0f-0b:
Correctly restore auto commit mode after exiting a transaction created via `DB.transaction`.

### DIFF
--- a/src/db.test.ts
+++ b/src/db.test.ts
@@ -334,12 +334,18 @@ Deno.test("transactions can be nested", function () {
 
   db.transaction(() => {
     db.query("INSERT INTO test (id) VALUES (1)");
-    try {
-      db.transaction(() => {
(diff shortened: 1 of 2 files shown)


dyedgreen:
Hey, thanks for the PR! Just to check: I updated the description above, does this correctly reflect your intend / the issue you’re trying to address with this change?

0f-0b:
Yes, it does. Thank you.

src/db.test.ts
```diff
@@ -341,6 +342,7 @@ Deno.test("transactions can be nested", function () {
     } catch (_) { /* ignore */ }
   });
 
+  assert(db.autoCommit);
```
dyedgreen, line 345: I think we should add `assert(!db.autoCommit)` calls inside of the transaction blocks as well.

dyedgreen (approved):
LGTM 🚀

PR #236, opened 2023-08-07
README: add kysely-deno-sqlite

alexgleason:
This library is used by [kysely-deno-sqlite](https://gitlab.com/soapbox-pub/kysely-deno-sqlite), which makes it possible to use SQLite + [Kysely](https://kysely.dev/) + Deno.


dyedgreen:
Thanks for making the PR!

dyedgreen:
(Also thanks for building the driver 🤩)

PR #238, opened 2023-08-13
add isClosed getter

### LINKED ISSUE #237: Expose the `#this.#open` property on the DB class instance
[Here](https://github.com/dyedgreen/deno-sqlite/blob/master/src/db.ts#L140) there seems to be a property that checks if the connection is still open. I would like to know if this can be exposed to the end-user.

My use-case would be due to [this](https://github.com/libsql/libsql-client-ts/issues/67) issue. I wasn't able to get the full API working but I would like to attempt to see if this package could be a viable option.

adoublef:
Closes #237 

This PR is to tackle [this issue](https://github.com/libsql/libsql-client-ts/issues/67), allowing me to make a Deno compatible Turso client that works locally and in a deployed environment.

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -135,602 +135,613 @@ export interface SqliteFunctionOptions {
  * queries.
  */
 export class DB {
-    #wasm: Wasm;
-    #functions: Array<(argc: number) => void>;
-    #open: boolean;
+  #wasm: Wasm;
+  #functions: Array<(argc: number) => void>;
+  #open: boolean;
 
-    #statements: Set<StatementPtr>;
-    #functionNames: Map<string, number>;
-    #transactionDepth: number;
+  #statements: Set<StatementPtr>;
+  #functionNames: Map<string, number>;
+  #transactionDepth: number;
 
-    /**
-     * Create a new database. The file at the
-     * given path will be opened with the
-     * mode specified in options. The default
-     * mode is `create`.
-     *
-     * If no path is given, or if the `memory`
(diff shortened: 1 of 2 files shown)


dyedgreen:
Overall this looks good! A few minor things:
- The diff seems to have a lot of white space changes, can you try to not re-format the whole file and only add the new function
- Can we add a doc comment, something to the effect of “Indicates if the database handle was previously closed. A closed database handle can no longer be used”
- Can you add some tests (this is pretty simple so something like create a db, assert it’s false, close it, assert it’s true, should be sufficient)

adoublef:
I have added a test and the jsdocs for the method. Regarding the formatting, I think this is a common problem with my IDE setup so thank you for pointing it out I need to fix it

src/db.ts
dyedgreen, line 1: The `isClosed` function looks good, however there are still lot's of unrelated formatting changes 👀

dyedgreen (changes requested):

src/db.ts
adoublef: I disabled auto format on my system as I think its just enforcing my coding style on every other project. hopefully that is sorted for you now

dyedgreen:
Thanks for bearing with me through the review, this looks good to me now! 🚀

dyedgreen (approved):

adoublef:
> Thanks for bearing with me through the review, this looks good to me now! 🚀 

No worries, thank you very much as now I think I can get this Turso client for Deno working as a POC!!

PR #251, opened 2023-08-31
Enable full-text search (FTS5)

alexgleason:
Fixes https://github.com/dyedgreen/deno-sqlite/issues/108 

cc @teleclimber


build/Makefile
```diff
@@ -33,6 +33,7 @@ CWRP = "./src/wrapper.c"
 # Configure sqlite for out use-case
 SQLFLG = -DSQLITE_DQS=0 -DSQLITE_THREADSAFE=0 -DSQLITE_LIKE_DOESNT_MATCH_BLOBS\
          -DSQLITE_DEFAULT_FOREIGN_KEYS=1 -DSQLITE_TEMP_STORE=2\
+         -DSQLITE_ENABLE_FTS5\
```
dyedgreen, line 36: Can we add the option at the end rather than in the middle and also add a note to the comment below why we’re adding it?

Otherwise this looks good to go!

dyedgreen (changes requested):

build/Makefile
alexgleason: I did that originally, then realized the options are in alphabetical order.

alexgleason: I broke the options onto separate lines, alphabetized them, and fixed/added comments for any missing ones.

dyedgreen (approved):
Awesome, thanks for cleaning this up! 🚀

PR #259, opened 2024-04-15
Fix deprecated APIs warnings

### LINKED ISSUE #255: Use of deprecated Deno API's
Starting with Deno 1.40, deno prints out a lot of warnings about the use of deprecated APIs in sqlite module. 

Below is a sample of the warnings (I removed some probably unnecessary wasm stack traces in order to condense the output a bit). 

```
warning: Use of deprecated "Deno.seekSync()" API. This API will be removed in Deno 2.
Stack trace:
  at js_write (https://deno.land/x/sqlite@v3.8/build/vfs.js:58:12)
  at <anonymous> (wasm://wasm/0028679a:1:5980)
hint: Use `file.seekSync()` instead.
hint: It appears this API is used by a remote dependency. Try upgrading to the latest version of that dependency.
warning: Use of deprecated "Deno.writeSync()" API. This API will be removed in Deno 2.
Stack trace:
  at js_write (https://deno.land/x/sqlite@v3.8/build/vfs.js:59:19)
  at <anonymous> (wasm://wasm/0028679a:1:5980)
hint: Use `writer.writeSync()` instead.
hint: It appears this API is used by a remote dependency. Try upgrading to the latest version of that dependency.
warning: Use of deprecated "Deno.seekSync()" API. This API will be removed in Deno 2.
Stack trace:
  at js_write (https://deno.land/x/sqlite@v3.8/build/vfs.js:58:12)
  at <anonymous> (wasm://wasm/0028679a:1:5980)
hint: Use `file.seekSync()` instead.
hint: It appears this API is used by a remote dependency. Try upgrading to the latest version of that dependency.
warning: Use of deprecated "Deno.writeSync()" API. This API will be removed in Deno 2.
Stack trace:
  at js_write (https://deno.land/x/sqlite@v3.8/build/vfs.js:59:19)
  at <anonymous> (wasm://wasm/0028679a:1:5980)
hint: Use `writer.writeSync()` instead.
```

canac:
Prepare for Deno 2.0 by migrating off of deprecated APIs. Fixes #255.

~~**Major Caveat**~~

~~With `DENO_FUTURE=1` set, there is no stable sync data operation available as of Deno v1.42. Enabling `DENO_FUTURE` removes access to the `Deno.fdatasyncSync` method, and `Deno.FsFile.syncDataSync` is still unstable. With `DENO_FUTURE=1` enabled, this library will only run with `--unstable-fs` (or `--unstable`).~~

~~Please let me know if you want me to document this caveat in the README or if you'd rather wait for `Deno.FsFile.syncDataSync` to stabilize before merging.~~

**Edit**: this limitation was resolved in Deno v1.44. This branch now successfully runs with `DENO_FUTURE=1`.

Thanks for creating this library! It's really easy to use in my scripts. I'm glad to have a chance to contribute back a little bit!

### DIFF
--- a/build/vfs.js
+++ b/build/vfs.js
@@ -87,13 +87,7 @@ export default function env(inst) {
     // Sync file data to disk
     js_sync: (rid) => {
       const file = getOpenFile(rid);
-      if (typeof file.rid === "number") {
-        // Uses deprecated Deno.FsFile.rid, which will be removed in Deno 2.0
-        Deno.fdatasyncSync(file.rid);
-      } else {
-        // Support DENO_FUTURE, but requires --unstable-fs
-        file.syncDataSync();
-      }
+      file.syncDataSync();
     },
     // Retrieve the size of the given file
     js_size: (rid) => {
--- a/build/vfs.js
+++ b/build/vfs.js
@@ -5,7 +5,7 @@ const isWindows = Deno.build.os === "windows";
 const OPEN_FILES = [];
 
 function getOpenFile(rid) {
-  const file = OPEN_FILES[rid]
+  const file = OPEN_FILES[rid];
   if (!file) {
     throw new Error(`Resource ID ${rid} does not exist.`);
   }
@@ -90,15 +90,15 @@ export default function env(inst) {
     },
     // Acquire a SHARED or EXCLUSIVE file lock
     js_lock: (rid, exclusive) => {
-      // this is unstable and has issues on Windows ...
-      if (Deno.flockSync && !isWindows) {
+      // this has issues on Windows ...
+      if (!isWindows) {
         getOpenFile(rid).lockSync(exclusive !== 0);
       }
     },
     // Release a file lock
     js_unlock: (rid) => {
-      // this is unstable and has issues on Windows ...
-      if (Deno.funlockSync && !isWindows) {
+      // this has issues on Windows ...
+      if (!isWindows) {
         getOpenFile(rid).unlockSync();
       }
     },
--- a/build/vfs.js
+++ b/build/vfs.js
@@ -2,6 +2,21 @@ import { getStr } from "../src/wasm.ts";
 
 const isWindows = Deno.build.os === "windows";
 
+const OPEN_FILES = new Map();
+
+function nextRid() {
+  const rid = (nextRid?.LAST_RID ?? 0) + 1;
+  nextRid.LAST_RID = rid;
+  return rid;
+}
+
+function getOpenFile(rid) {
+  if (!OPEN_FILES.has(rid)) {
+    throw new Error(`Resource ID ${rid} does not exist.`);
+  }
+  return OPEN_FILES.get(rid);
+}
+
 // Closure to return an environment that links
 // the current wasm context
 export default function env(inst) {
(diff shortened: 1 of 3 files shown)


dyedgreen:
Hello, thanks for the contribution!

This mostly looks good code wise, but requiring unstable to run is not great. I’ll see what the status of these APIs is on the deno side 👀

canac:
I pushed a new commit that uses the not-yet stabilized APIs. The tests are failing now, but they should pass if we rerun them after Deno v1.44 releases.

canac:
Tests pass now on Deno v1.44! Let me know if you have any suggestions or feedback on this PR.

build/vfs.js
```diff
@@ -55,30 +75,38 @@ export default function env(inst) {
...
     js_size: (rid) => {
-      return Deno.fstatSync(rid).size;
+      const file = getOpenFile(rid);
+      return file.statSync().size;
     },
     // Acquire a SHARED or EXCLUSIVE file lock
     js_lock: (rid, exclusive) => {
       // this is unstable and has issues on Windows ...
-      if (Deno.flockSync && !isWindows) Deno.flockSync(rid, exclusive !== 0);
+      if (Deno.flockSync && !isWindows) {
```
canac, line 100: @dyedgreen What do you think is the best way to detect the availability of `FsFile.lockSync`?

* `FsFile.lockSync` is available in v1.44.0, but `Deno.flockSync` is still undefined without `--unstable` or `--unstable-fs`. So we can't check `Deno.flockSync`.
* `FsFile.lockSync` exists in versions before v1.44.0, but calling it terminates deno, even from inside of a try/catch block. So we can't check the existence of `FsFile.lockSync`.
* `Deno.version.deno >= "1.44.0"` is the only condition I could get to work. Should we go with that?

dyedgreen: Yeah I feel like we’ll just need to do a version check here 😕

build/vfs.js
```diff
@@ -55,30 +75,38 @@ export default function env(inst) {
...
     js_truncate: (rid, size) => {
-      Deno.ftruncateSync(rid, size);
+      const file = getOpenFile(rid);
+      file.truncateSync(size);
     },
     // Sync file data to disk
     js_sync: (rid) => {
-      Deno.fdatasyncSync(rid);
+      const file = getOpenFile(rid);
+      file.syncDataSync();
```
canac, line 90: @dyedgreen I don't know of a good way to make this method support Deno versions before v1.44.

* We could make it a no-op like the lock/unlock method. Is that acceptable or would that break things?
* Or we could bump the minimum supported Deno version to v1.44 after this lands.

What do you think?

canac: @dyedgreen Let me know when you get a chance to look at this. I'm thinking we may need to bump the minimum supported Deno version to get access to the new file flushing APIs. What are your thoughts?

canac:
@dyedgreen With Deno 2.0 supposedly right around the corner, it would be awesome to push this PR across the finish line. My one remaining question is whether bumping up the minimum support Deno version to v1.44 is acceptable. Cheers!

dyedgreen:
Hey! This mostly looks good. One small thing I'd want to change is use an array instead of a map to store the open files; I made some changes locally but I can't seem to push them to your branch -- can you take a look at making that change?

Otherwise this LGTM!

canac:
@dyedgreen I checked the box, but sorry it won't let you push to the branch. Thanks for looking at it again! I'll try to make that change this week.

canac:
I tested this branch with a bunch of Deno versions, and it works in Deno pre-v1.44 as long as you pass in the `--unstable-fs` flag.

dyedgreen (approved):
LGTM

PR #260, opened 2024-04-20
Fix test failing because of syntax error

### LINKED ISSUE #256: Assertion in db.test.ts succeeds for the wrong reason
I'm wondering if the typo in this SQL statement is intended:
https://github.com/dyedgreen/deno-sqlite/blob/1e98e837c6b2efe1f7b6291501bbe49aca296c9e/src/db.test.ts#L513

Shouldn't the statement fail because of read-only access rather than a syntax error?

canac:
Fixes #256


dyedgreen (approved):

dyedgreen:
LGTM

Thank you for taking the time to fix this!

PR #262, opened 2024-05-07
chore: publish to jsr

danewalters:


danewalters:
This is my first time submitting a PR to an open source repository, so I may not have done a very good job.
I have two ideas.
- Would it be better to directly maintain a jsr.json file?
- Would it be better to exclude test files from the jsr.json file?

PR #268, opened 2024-10-27
Update Browser Build for Deno 2

aunyks:
## TL;DR
- Uses [esbuild-deno-loader](https://jsr.io/@luca/esbuild-deno-loader) as [a replacement](https://stackoverflow.com/questions/76028937/deno-bundle-replacement) for `deno bundle` ([now deprecated](https://deno.com/blog/v2.0-release-candidate#command-line-interface-changes))
- Updates `vfs.js`, which supports the browser build, to be in line with the local build
- Uses `deno.json` to handle import maps ([standard in Deno 2](https://docs.deno.com/runtime/fundamentals/configuration/#dependencies)) for the browser build
- Updates `README.md` in line with the above changes

## Notes for Reviewers
- The formatting changes in `vfs.js` arise from running `deno fmt` against the file

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ There is **experimental** support for using `deno-sqlite` in the browser. You
 can generate a browser compatible module by running:
 
 ```bash
-deno bundle --import-map browser/import_map.json browser/mod.ts [output_bundle_path]
+deno run --allow-run --allow-env --allow-read --allow-write ./browser/bundle.js
 ```
 
 The modules documentation can be seen by running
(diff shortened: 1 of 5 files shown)


dyedgreen:
Hey, sorry for taking a while to get back to this!

This change looks great, outside of a few files that are flagged by the formatting in CI 👀

(Some of those have not changed in a while so they might have changed the formatting in Deno?)

README.md
```diff
@@ -85,7 +85,7 @@ There is **experimental** support for using `deno-sqlite` in the browser. You
 can generate a browser compatible module by running:
 
 ```bash
-deno bundle --import-map browser/import_map.json browser/mod.ts [output_bundle_path]
+deno run --allow-run --allow-env --allow-read --allow-write ./browser/bundle.js
```
dyedgreen, line 88: I think we can look down permissions a bit more here?

`--allow-run` isn’t required, and we can scope reading to the repo and writing to just the bundle.

PR #272, opened 2026-07-16
Fix database locks forever when db close not called from reader

### LINKED ISSUE #271: Cannot access the same database from a Worker
_I had previously misunderstood the cause of this issue._

When opening a database from two different isolates, either two different workers or a worker and the main isolate, and performing operations on both, results in undefined behavior. In the reproduction case below, I am writing to the database from the main and reading once from the database from a child isolate. Doing so consistently freezes the main.

## Repro

```TS
import { DB } from "https://deno.land/x/sqlite@v3.9.1/mod.ts";

const url = new URL(import.meta.url);
const dbPath = import.meta.resolve("./repro.sqlite3");

if (!url.searchParams.has("worker")) {
  const db = new DB(dbPath, { mode: "create", uri: true });
  db.query("CREATE TABLE IF NOT EXISTS rows (id INTEGER PRIMARY KEY)");
  db.execute("DELETE FROM rows");

  let id = 1;
  setInterval(() => {
    console.log(`INSERT ${id}`);
    db.query("INSERT INTO rows (id) VALUES (?)", [id]);
    id++;
  }, 50);

  setTimeout(() => {
    const workerUrl = new URL(import.meta.url);
    workerUrl.searchParams.set("worker", "1");
    new Worker(workerUrl, { type: "module" });
  }, 1000);
} else {
  const db = new DB(dbPath, { uri: true });
  console.log("Reading...");
  const result = db.query("SELECT id FROM rows ORDER BY id DESC LIMIT 1");
  console.log(`SELECTED ${result[0][0]}`);
}
```

```
deno run -A repro.ts 
INSERT 1
INSERT 2
...
INSERT 19
INSERT 20
Reading...
SELECTED 20
INSERT 21
```

As soon as the worker does a select, the main isolate freezes on the next write, unable to commit.

## vfs

What is have also tried:

`const dbPath = import.meta.resolve("./repro.sqlite3") + "?vfs=...";`
But none of the values from this list work:

- unix
- unix-dotfile
- unix-excl
- unix-none
- unix-posix

```
error: Uncaught (in promise) SqliteError: no such vfs: unix-flock
      throw new SqliteError(this.#wasm, status);
```

lexidor:
fixes #271

The `denoUnlock` function did not correctly release a lock given the argument `SQLITE_LOCK_NONE`.
This function is supposed to release locks when given `_NONE`.
This function is supposed to reduce lock strength from high to SHARED when given `_SHARED`.

From sqlite3.c:L1077-L1079

```
** xUnlock() downgrades the database file lock to either SHARED or NONE.
*  If the lock is already at or below the requested lock state, then the call
** to xUnlock() is a no-op.
```

I am not sure about the "xUnlock() is a no-op".
It sound like sqlite expects us to do:

```
// `p` is the DenoFile pointer and eLock is desired level.
if (p->currentLogLevel < eLock) {
  return;
}
```

I did not find a way to query the current lock level of the file.

---

Before recompiling with make, the following (new) test fails:

```
 ERRORS 

worker read releases database lock => ./src/db.test.ts:443:6
error: AssertionError: Values are not equal:

    [Diff] Actual / Expected

-   timeout
+   wrote

  throw new AssertionError(message);
```


lexidor:
I did not include the compiled assets.
Reviewing that for malicious changes is neigh impossible.
Do not forget to compile the new wasm. :slightly_smiling_face:

dyedgreen (approved):
This looks reasonable. Thanks for the fix!

